Issue #25, opened 2025-08-12
Onboarding wizard


positonic:
I really like the onboarding screen on get them.io where it asks the user what apps that they use. We could consider do this as a way for a new user to onboard and create workflows based on what tools that they use.

<img width="1719" height="942" alt="Image" src="https://github.com/user-attachments/assets/7f6b80a5-f22d-45b0-a0bb-e73f74b8d472" />

<img width="1714" height="934" alt="Image" src="https://github.com/user-attachments/assets/450f016b-82ab-42b2-ba12-10f9c9f613f7" />

positonic:
I think we've made progress on this, but it's not quite finished yet. We should discuss what exactly should happen for this.

PR #26, opened 2025-08-12
Add: user onboarding flow

Prajjawalk:
Added user onboarding flow corresponding to issue #25

### DIFF
--- a/.taskmaster/docs/user-onboarding-prd.md
+++ b/.taskmaster/docs/user-onboarding-prd.md
@@ -1,0 +1,300 @@
+# User Onboarding Flow PRD
+
+## Project Overview
+
+**Project Name:** Exponential User Onboarding Flow  
+**Priority:** High  
+**Timeline:** 2-3 weeks  
+**Version:** 1.0  
+
+## Executive Summary
+
+Create a comprehensive user onboarding flow for the Exponential productivity management application that guides new users through account setup, usage preference selection, role specification, tool integration, and first project creation to ensure successful user activation and engagement.
+
+## Problem Statement
+
+Currently, new users who sign up for Exponential may feel overwhelmed by the full feature set without proper guidance on how to get started. This leads to:
+- Low user activation rates
+- Confusion about which features to use first
+- Poor initial user experience
+- Higher abandonment rates during first session
+
+## Solution Overview
+
+Implement a guided 4-step onboarding flow that:
+1. Determines user intent (work vs personal)
+2. Captures role information (if work-related)
+3. Enables tool integrations based on user preferences
+4. Creates the user's first project to demonstrate core functionality
+
+## User Stories
+
+### Primary User Stories
+
+**US-001:** As a new user, I want to specify whether I'm using Exponential for work or personal purposes so that the application can tailor the experience to my needs.
+
+**US-002:** As a work user, I want to specify my role in the company so that the application can suggest relevant features and workflows.
+
+**US-003:** As a new user, I want to select which tools I currently use so that Exponential can integrate with my existing workflow.
+
+**US-004:** As a new user, I want to create my first project during onboarding so that I can immediately start using the application productively.
+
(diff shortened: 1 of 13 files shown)


positonic:
## Build Fix Summary

Hey @Prajjawalk\! I've pushed a fix for the build errors. Here's what was causing the issues and how to avoid them in the future:

### What went wrong

The build was failing due to TypeScript type mismatches, not webpack errors. Here were the three issues:

#### 1. **Priority Enum Case Mismatch** 
- **Problem**: The  was using lowercase priority values (`'low' | 'medium' | 'high'`) while the API expected uppercase (`'LOW' | 'MEDIUM' | 'HIGH'`)
- **Fix**: Updated all priority values to uppercase in the component
- **Future tip**: Always check the API schema/types when working with enums. The tRPC router defined it as `z.enum(["LOW", "MEDIUM", "HIGH"])`

#### 2. **Invalid Mantine Component Prop**
- **Problem**: The `breakpoint` prop was used on the Stepper component, but this prop doesn't exist in Mantine v7
- **Fix**: Removed the invalid prop
- **Future tip**: When using UI library components, refer to their documentation for valid props. Mantine's docs are at https://mantine.dev/

#### 3. **Missing User Fields in Query**
- **Problem**: The WhatsApp service was querying user data but the select statement was missing the new onboarding fields you added
- **Fix**: Added the missing fields (usageType, userRole, selectedTools, onboardingCompletedAt, onboardingStep)
- **Future tip**: When adding new fields to the database schema, search the codebase for places that query that table to ensure all necessary queries are updated

### How to catch these issues locally

Before pushing, always run:
```bash
npm run build
npm run typecheck
npm run lint
```

These commands will catch TypeScript errors before they hit CI/CD.

### Pro tip
The project has a pre-push hook that runs these checks automatically. If you ever need to bypass it temporarily (not recommended), you can use `git push --no-verify`, but it's better to fix the issues first.

Great work on implementing the onboarding flow\! The feature looks solid, just needed these small type fixes. 🚀

positonic:
It would also have saved me time if you could have told me how to test this directly. I think this is best practice when you're adding a description of a PR. Please, let's do this in the future.

positonic:
## Build Error Solution

I see you're still getting build errors. Here's what's happening and how to fix it:

### Issue 1: Missing @openai/realtime-api-beta
```
Module not found: Can't resolve '@openai/realtime-api-beta'
```

This package uses a GitHub URL in package.json. You need to:

1. **Ensure you have GitHub access configured:**
```bash
# Check if you can access GitHub packages
git ls-remote https://github.com/simliai/openai-realtime-api-beta.git
```

2. **Clean install with bun:**
```bash
rm -rf node_modules bun.lockb
bun install --force
```

### Issue 2: Webpack Loader JSON Errors

The `Unexpected end of JSON input` errors with `middlewareConfig=e30%3D` indicate corrupted build cache or filesystem issues.

**Complete fix sequence:**

```bash
# 1. Clean EVERYTHING
rm -rf node_modules .next bun.lockb package-lock.json

# 2. Clear bun cache
bun pm cache rm

# 3. Fresh install
bun install

# 4. Clear Next.js cache and rebuild
rm -rf .next
SKIP_ENV_VALIDATION=1 bun run build
```

### If that doesn't work:

**Check for filesystem issues:**
```bash
# Check disk space
df -h .

# Check for file permission issues
ls -la src/app/api/

# Try with npm instead of bun
rm -rf node_modules bun.lockb
npm install
npm run build
```

### Root cause possibilities:
1. **Corrupted bun cache** - The JSON parsing errors suggest corrupted cached data
2. **Filesystem permissions** - Check if your user has full write access
3. **Disk space** - Low disk space can cause partial writes
4. **Antivirus interference** - Some AV software can interfere with node_modules

### Nuclear option:
If nothing else works, clone fresh:
```bash
cd ..
git clone https://github.com/positonic/exponential exponential-fresh
cd exponential-fresh
bun install
bun run build
```

Let me know which error persists after trying these steps\!

PR #280, opened 2026-06-23
feat(crm-api): dedupe contactCreate by emailHash, add email filter to contactList

Prajjawalk:
## Summary

Public CRM API (`crmApi.*`) machine-intake hardening. Two related changes bundled:

1. **Workspace-scoped uniqueness on `CrmContact.emailHash`.** Was globally `@unique`. Switched to `@@unique([workspaceId, emailHash])` so the same person can legitimately be a contact in more than one workspace — and so workspace B isn't blocked because workspace A already claimed the email.
2. **Idempotent `contactCreate` even under concurrent double-submits.** Pre-check + insert is a check-then-act race; the loser of two concurrent requests was hitting `P2002 → 500`. Insert is now wrapped in a `try/catch` that re-fetches the winner on `P2002` and returns it as a dedupe hit.

### Behaviour

- `contactCreate` looks up by `(workspaceId, emailHash)`. Same-workspace hit returns the existing row untouched with `wasExisting: true`. Cross-workspace hit no longer surfaces (workspace-scoped lookup) — the call lands a fresh row for *this* workspace with `wasExisting: false`. Concurrent double-submit within a workspace → `P2002` is caught, winner re-fetched, returned with `wasExisting: true`.
- `contactList` accepts an optional `email` filter resolved through the same composite index — lets integrations check "is this applicant already in *our* queue?" before submitting.
- Response shape: `wasExisting: boolean` on `contactCreate` lets callers distinguish dedup hits from fresh inserts.

The session-protected `crmContact.create` is intentionally left as-is (in-app manual entry, still allows duplicate emails per its docstring). `verifyWorkspaceAccess` semantics are unchanged.

## Files

- `prisma/schema.prisma` — drop global `@unique` on `emailHash`, add `@@unique([workspaceId, emailHash])`
- `prisma/migrations/20260624000000_crm_contact_email_unique_per_workspace/migration.sql` — drop the old unique index, create the composite one
- `src/server/api/routers/crmApi.ts` — composite lookup; drop cross-workspace CONFLICT; P2002 catch + refetch; new `email` filter on `contactList`; `wasExisting` on `contactCreate` response
- `src/server/services/crm/createCrmContact.ts` — composite lookup; drop cross-workspace error; P2002 catch + refetch (Forms-intake idempotency under race)
- `src/server/api/routers/transcription.ts` — two `findUnique({ where: { emailHash } })` sites switched to composite; drop the free-text path's cross-workspace skip branch (workspace-scoped lookup makes it moot)
- `chore(deps)` commit: install `@anthropic-ai/sdk` (declared in `package.json`, missing from the lockfile, blocked unit tests)

## Migration prerequisite

This PR adds a schema migration. **It must be applied before deploy** — the new code path uses the composite unique key.

```bash
npx prisma migrate dev      # local
# Production: applied via your standard migrate-deploy step before this PR is rolled out
```

The migration is data-safe: the prior global constraint was strictly enforced, so no existing rows collide on the new composite.

## Test plan

- [x] `npm test -- crmApi` — 7/7 pass
  - [x] new-email → fresh row + emailHash, `wasExisting: false`
  - [x] same-workspace email hit → existing row, `wasExisting: true`, no `create()` call, composite key in lookup
  - [x] cross-workspace email → fresh row in this workspace, `wasExisting: false`
  - [x] P2002 race → re-fetch winner, return with `wasExisting: true`
  - [x] no email → create without emailHash, no dedup lookup
  - [x] `contactList` with `email` → adds emailHash to where
  - [x] `contactList` without `email` → omits emailHash from where
- [x] Full repo unit suite — 1076/1076 pass (pre-commit hook)
- [x] `npx tsc --noEmit` — clean
- [x] Prisma migration check (pre-commit hook) — pass
- [x] Hardcoded-colors check (pre-commit hook) — pass

## Reviewer notes

- **Pre-existing contacts created via the old API path** have `emailHash = NULL` and will not match the dedup lookup or the `contactList` email filter. Acceptable for v1 (eventual consistency); a follow-up backfill migration can be done later if you want them to participate in dedup.
- The `transcription.ts` free-text path used to deliberately skip linking a participant to a cross-workspace CRM contact ("rather than leaking across the workspace boundary"). With workspace-scoped uniqueness, the lookup is already scoped to this workspace, so the skip branch became dead code and was removed.
- Addresses Asks 1 + 2 from the integration asks doc (CLEAR waitlist integration). Asks-the-CLEAR-team-doesn't-care-about (Ask 3, `CrmOrganization` procedures under `crmApi.*`) is intentionally not in scope.

## Summary by CodeRabbit

* **New Features**
  * Added optional email-based filtering to CRM contact listings.
  * Updated contact creation to be idempotent using a workspace-scoped email hash to prevent duplicates within the same workspace.
* **Bug Fixes**
  * Handle concurrent “create” attempts by returning the already-created contact when a uniqueness conflict occurs.
  * Adjusted meeting participant linking so email-to-contact resolution is scoped to the same workspace.
* **Tests**
  * Added unit tests covering contact deduplication, email filtering, and concurrent create/race-condition behavior.

### DIFF
--- a/src/server/api/routers/crmApi.ts
+++ b/src/server/api/routers/crmApi.ts
@@ -5,6 +5,7 @@ import { apiKeyMiddleware } from "~/server/api/middleware/apiKeyAuth";
 import { encryptString, decryptBuffer } from "~/server/utils/encryption";
 import type { CrmContact, PrismaClient } from "@prisma/client";
 import { getProjectAccess, hasProjectAccess } from "~/server/services/access";
+import { emailHashFor } from "~/server/services/crm/createCrmContact";
 
 // ─── Helpers ────────────────────────────────────────────────────────
 
@@ -146,6 +147,7 @@ export const crmApiRouter = createTRPCRouter({
       z.object({
         workspaceId: z.string(),
         search: z.string().optional(),
+        email: z.string().email().optional(),
         tags: z.array(z.string()).optional(),
         organizationId: z.string().optional(),
         limit: z.number().min(1).max(100).optional(),
@@ -153,12 +155,17 @@ export const crmApiRouter = createTRPCRouter({
       }),
     )
     .query(async ({ ctx, input }) => {
-      const { workspaceId, search, tags, organizationId, limit = 50, cursor } = input;
+      const { workspaceId, search, email, tags, organizationId, limit = 50, cursor } = input;
       await verifyWorkspaceAccess(ctx.db, workspaceId, ctx.userId);
+
+      // When `email` is supplied, dedupe lookup via the globally-unique emailHash.
+      // Contacts created before emailHash was set on the public API will not match.
+      const emailHash = email ? emailHashFor(email) : undefined;
 
       const contacts = await ctx.db.crmContact.findMany({
         where: {
           workspaceId,
+          ...(emailHash ? { emailHash } : {}),
           ...(search
             ? {
                 OR: [
@@ -256,6 +263,43 @@ export const crmApiRouter = createTRPCRouter({
         }
       }
 
+      const contactInclude = {
+        organization: true,
+        createdBy: {
+          select: { id: true, name: true, email: true, image: true },
+        },
+      } as const;
+
+      // Dedupe by globally-unique emailHash. The create path is the canonical
+      // machine intake; an applicant submitting the same form twice should land
+      // one row, not many. Existing rows are returned untouched (no field
+      // overwrite) so callers can distinguish via `wasExisting`.
+      const trimmedEmail = contactData.email?.trim();
+      const emailHash = trimmedEmail ? emailHashFor(trimmedEmail) : null;
+
+      if (emailHash) {
+        const existing = await ctx.db.crmContact.findUnique({
+          where: { emailHash },
+          select: { id: true, workspaceId: true },
+        });
+        if (existing) {
+          if (existing.workspaceId !== workspaceId) {
+            // emailHash is globally unique. Returning a contact owned by a
+            // different workspace would leak a cross-tenant reference; surface
+            // a clear conflict instead.
+            throw new TRPCError({
+              code: "CONFLICT",
+              message: "A contact with this email already exists in another workspace",
+            });
+          }
+          const existingContact = await ctx.db.crmContact.findUniqueOrThrow({
+            where: { id: existing.id },
+            include: contactInclude,
+          });
+          return { ...safeDecryptContact(existingContact), wasExisting: true };
+        }
+      }
+
       const dbData: Record<string, unknown> = {
         workspaceId,
         createdById: ctx.userId,
@@ -266,6 +310,7 @@ export const crmApiRouter = createTRPCRouter({
         about: contactData.about ?? undefined,
         profileType: contactData.profileType ?? undefined,
         organizationId: contactData.organizationId ?? undefined,
+        ...(emailHash ? { emailHash } : {}),
       };
 
       for (const field of piiFields) {
@@ -275,15 +320,10 @@ export const crmApiRouter = createTRPCRouter({
 
       const contact = await ctx.db.crmContact.create({
         data: dbData as Parameters<typeof ctx.db.crmContact.create>[0]["data"],
-        include: {
-          organization: true,
-          createdBy: {
-            select: { id: true, name: true, email: true, image: true },
-          },
-        },
+        include: contactInclude,
       });
 
-      return safeDecryptContact(contact);
+      return { ...safeDecryptContact(contact), wasExisting: false };
     }),
 
   contactUpdate: apiKeyMiddleware
--- a/src/server/api/routers/crmApi.ts
+++ b/src/server/api/routers/crmApi.ts
@@ -3,6 +3,7 @@ import { createTRPCRouter } from "~/server/api/trpc";
 import { TRPCError } from "@trpc/server";
 import { apiKeyMiddleware } from "~/server/api/middleware/apiKeyAuth";
 import { encryptString, decryptBuffer } from "~/server/utils/encryption";
+import { Prisma } from "@prisma/client";
 import type { CrmContact, PrismaClient } from "@prisma/client";
 import { getProjectAccess, hasProjectAccess } from "~/server/services/access";
 import { emailHashFor } from "~/server/services/crm/createCrmContact";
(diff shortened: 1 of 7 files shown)


src/server/api/routers/crmApi.ts
```diff
@@ -256,6 +263,43 @@ export const crmApiRouter = createTRPCRouter({
...
+
+      // Dedupe by globally-unique emailHash. The create path is the canonical
+      // machine intake; an applicant submitting the same form twice should land
+      // one row, not many. Existing rows are returned untouched (no field
+      // overwrite) so callers can distinguish via `wasExisting`.
+      const trimmedEmail = contactData.email?.trim();
+      const emailHash = trimmedEmail ? emailHashFor(trimmedEmail) : null;
+
+      if (emailHash) {
+        const existing = await ctx.db.crmContact.findUnique({
```
positonic, line 283: **[Medium] Dedupe `findUnique`→`create` is not atomic — concurrent identical submits 500**

The `findUnique(emailHash)` check and the later `create` are separate queries with no transaction or unique-violation handling. The exact scenario this feature targets — "an applicant submitting the same form twice" — is usually a near-simultaneous double-submit (double-click / retry). Both requests find no existing row, both proceed to `create`, and the second hits the globally-unique `emailHash` constraint and throws Prisma `P2002`, which surfaces as a `500 INTERNAL_SERVER_ERROR` rather than the intended deduped `200 { wasExisting: true }`.

Suggested fix: catch the unique-violation and re-resolve to the existing row (same outcome as a sequential dedupe hit):

```ts
try {
  const contact = await ctx.db.crmContact.create({ data: ..., include: contactInclude });
  return { ...safeDecryptContact(contact), wasExisting: false };
} catch (e) {
  if (e instanceof Prisma.PrismaClientKnownRequestError && e.code === "P2002" && emailHash) {
    const existingContact = await ctx.db.crmContact.findUniqueOrThrow({
      where: { emailHash }, include: contactInclude,
    });
    if (existingContact.workspaceId !== workspaceId) {
      throw new TRPCError({ code: "CONFLICT", message: "A contact with this email already exists in another workspace" });
    }
    return { ...safeDecryptContact(existingContact), wasExisting: true };
  }
  throw e;
}
```

<details>

In `src/server/api/routers/crmApi.ts`, the `contactCreate` mutation dedupes by reading `crmContact.findUnique({ where: { emailHash } })` and, if absent, calling `crmContact.create(...)`. `emailHash` has a global UNIQUE constraint. Two concurrent requests with the same email both pass the check and both attempt the insert; the loser throws Prisma `P2002`, which becomes a 500 — defeating the dedupe in the precise double-submit case the feature exists for.

Make the create resilient: wrap `crmContact.create` in a try/catch, and on `Prisma.PrismaClientKnownRequestError` with `code === "P2002"`, re-query by `emailHash`, apply the same cross-workspace CONFLICT guard, and return the existing row with `wasExisting: true`. Import `Prisma` from `@prisma/client`. Keep the existing pre-check (it still avoids the wasted insert in the common sequential case). Don't change the mutation's input/output shape. Add a regression test that simulates `create` rejecting with a P2002 and asserts the mutation resolves to `{ wasExisting: true }` rather than throwing.

</details>

src/server/api/routers/crmApi.ts
```diff
@@ -256,6 +263,43 @@ export const crmApiRouter = createTRPCRouter({
...
+          select: { id: true, workspaceId: true },
+        });
+        if (existing) {
+          if (existing.workspaceId !== workspaceId) {
+            // emailHash is globally unique. Returning a contact owned by a
+            // different workspace would leak a cross-tenant reference; surface
+            // a clear conflict instead.
+            throw new TRPCError({
+              code: "CONFLICT",
+              message: "A contact with this email already exists in another workspace",
```
positonic: **[Low] Cross-workspace CONFLICT lets an API-key holder probe emails across tenants**

`emailHash` is globally unique, so this branch fires whenever the email exists in *any other* workspace. An API key is workspace-scoped, but this turns `contactCreate` into a cross-tenant existence oracle: a holder of workspace A's key can submit arbitrary emails and learn — from the explicit `"already exists in another workspace"` message — whether each one exists somewhere else on the platform. For a CRM/waitlist of named individuals that's a meaningful disclosure.

This is partly inherent to the global-unique design (the old code would also throw a P2002 on collision), but the explicit message makes the signal cleaner and unambiguous. Consider a generic message that doesn't confirm *another workspace* specifically:

```suggestion
              message: "A contact with this email already exists",
```

<details>

In `src/server/api/routers/crmApi.ts`, `contactCreate` throws `CONFLICT` with the message "A contact with this email already exists in another workspace" when an `emailHash` collision is owned by a different workspace. Because `emailHash` is globally unique and API keys are workspace-scoped, this is a cross-tenant email-existence oracle, and the message explicitly confirms the email lives in another tenant.

Soften the disclosure: keep the CONFLICT status (callers still need to know the create didn't succeed) but make the user-facing message generic ("A contact with this email already exists") so it doesn't confirm cross-tenant existence. Keep the precise reason in a server-side log if useful for debugging. Don't change the status code or the same-workspace dedupe behavior.

</details>

positonic (approved):
Staff-level review — feature logic is solid and mirrors the existing `createCrmContact` service cleanly. Approving; the findings below are non-blocking follow-ups (1 should ideally be addressed before merge).

**Summary:** 0 Critical, 0 High, 2 Medium, 1 Low.

---

### [Medium] The `bun.lock` change is moot — drop it

The commit `chore(deps): install missing @anthropic-ai/sdk` justifies the lockfile change as "blocking the unit test runner." But this repo is npm-based: CI runs `npm ci` against `package-lock.json` (`.github/workflows/test.yml`), and `@anthropic-ai/sdk@^0.104.2` is **already present** in `package-lock.json` (it landed in #268). Nothing in CI or the test runner reads `bun.lock`.

What this commit actually does is regenerate a stale `bun.lock` via `bun install`, pulling in ~400 lines of unrelated drift (`@sentry/nextjs`, `@xyflow/react`, `@dagrejs/dagre`, several `@tiptap/*` extensions) that have nothing to do with this PR — a large block of unreviewed dependency-graph churn from the wrong package manager.

**Suggested fix:** drop the `bun.lock` commit from this PR (`git restore --source main bun.lock` or rebase it out) so the PR contains only the feature + tests. If `bun.lock` is genuinely unused, remove it from the repo in a separate cleanup PR so it stops drifting.

<details>

In PR #280, one commit modifies `bun.lock` (+402/-33 lines), justified as installing a missing `@anthropic-ai/sdk` that was "blocking the unit test runner." This is incorrect: the repo uses npm (`package-lock.json`, CI runs `npm ci` per `.github/workflows/test.yml`), and `@anthropic-ai/sdk` is already in `package-lock.json`. The `bun.lock` edit is irrelevant churn that also pulls in many unrelated packages.

Remove the `bun.lock` change from this PR so it contains only the feature + tests (`crmApi.ts`, `crmApi.test.ts`). Verify `npm ci && npm test -- crmApi` still passes without touching `bun.lock`. Do not modify `package-lock.json`. If you want to address the stale `bun.lock` separately, do that in its own cleanup PR.

</details>

PR #438, opened 2026-07-30
feat(ai-worker): run Claude Code directly against OpenRouter, add ai-buildable

positonic:
Swaps the autonomous ticket worker onto OpenRouter (default `z-ai/glm-5.2`) and adds a second trigger label for features. **No app code, no schema change, no stored credentials.**

Amends [ADR-0029](docs/adr/0029-ai-bug-fixer-workflow.md) rather than superseding it — the shape of that decision (Actions, CLI-as-interface, PR for human review) is unchanged.

## 1. Claude Code CLI called directly, not `claude-code-action`

The action can't target a third-party endpoint, which the model change requires. It also rewrote `remote.origin.url` to embed its own bot token, which the workflow had to undo before every push. Decisively: its step is `continue-on-error`, so failures surfaced as a **green run that produced no PR** — a total outage went unnoticed for five weeks. A direct call gives a plain exit code.

Drops `id-token: write` and the remote-URL workaround. The agent is allow-listed to `Read,Write,Edit,Glob,Grep` and **Bash is explicitly denied** via `--disallowedTools` (an allow-list alone doesn't restrict — the checked-in `.claude/settings.json` would merge its own `Bash(...)` allow rules in), so an agent holding an API key in a public repo's runner can't shell out. Output goes to a file, never the log, with the tail surfaced on failure.

## 2. OpenRouter, default `z-ai/glm-5.2`, for bugs and features alike

Via OpenRouter's Anthropic-compatible endpoint, so "which provider" collapses to a model string. `AI_BUILDER_BASE_URL` can be set to `https://api.anthropic.com` to talk to Anthropic directly (clearing it just restores the OpenRouter default — Actions `||` can't tell unset from empty).

**Billing is now metered, not a flat subscription fee** — that was listed as a cost control in ADR-0029 and is genuinely gone. `--max-budget-usd` (default $2/run) was added as a hard ceiling to compensate.

`AI_BUG_FIXER_MODEL` is deliberately no longer read: it named an Anthropic model, and a stale value would be sent to an endpoint that doesn't know it.

## 3. `ai-buildable` as a second trigger label

`--label` ANDs, so "either label" can't be one query — the scan runs one list per label and `select-candidate.mjs` merges them, recording which matched. A ticket with both is worked as `ai-fixable` (stricter profile).

`ai-buildable` adds one gate: **schema and migration edits are rejected by diff, not by prompt.** Preview deploys run `next build` only, never `prisma migrate deploy`, so a schema change can't reach a preview database — the feature would look broken for reasons unrelated to the agent.

## Required before merge

```bash
gh secret set AI_BUILDER_API_KEY -R positonic/exponential   # OpenRouter key
```

The `ai-buildable` label has already been created in `syntrofi`. This was load-bearing: the CLI errors on an unknown label slug and the scan doesn't swallow errors, so merging without it would fail the job hourly.

Optional cleanup — nothing reads these any more:
```bash
gh secret delete CLAUDE_CODE_OAUTH_TOKEN -R positonic/exponential
```

## Testing

- `select-candidate.mjs` exercised over merge, both-labels dedupe, security exclusion, and PR cap
- Workflow YAML parsed; step order and job outputs confirmed
- `--max-turns` verified still valid despite being undocumented in 2.1.186 (unknown flags are rejected instantly; this one isn't). `--max-budget-usd` confirmed present.

**Not yet verified:** that GLM 5.2 actually drives Claude Code's tool loop end to end — no OpenRouter key was available. That's a 30-second check once the secret is set, and only the model default depends on it; the runner change stands either way.

### DIFF
--- a/.github/workflows/ai-bug-fixer.yml
+++ b/.github/workflows/ai-bug-fixer.yml
@@ -246,26 +246,35 @@ jobs:
         id: agent
         continue-on-error: true
         env:
-          # ANTHROPIC_AUTH_TOKEN (not ANTHROPIC_API_KEY) is the header the SDK
-          # sends to a custom base URL. ANTHROPIC_API_KEY is blanked explicitly
-          # so a leftover Anthropic key in repo secrets cannot win the race and
-          # silently bill the wrong account.
+          # Exactly ONE credential reaches the agent (AI_BUILDER_API_KEY); the
+          # run block exports it under the variable whose auth header the
+          # target endpoint expects. Anthropic direct authenticates with
+          # x-api-key, which the SDK sends from ANTHROPIC_API_KEY; every other
+          # Anthropic-compatible endpoint (OpenRouter included) takes
+          # Authorization: Bearer, sent from ANTHROPIC_AUTH_TOKEN. Both are
+          # pre-blanked here so a leftover key in repo settings can never win a
+          # precedence race and silently bill the wrong account.
           #
           # That precedence trap is not hypothetical: PR #436 traced a five-week
           # outage to claude-code-action PREFERRING claude_code_oauth_token when
           # both credentials were present. The subscription behind it had hit a
           # rolling usage cap and failed in ~250ms, and adding API-key credit
-          # changed nothing because the API key was never consulted. Exactly one
-          # credential reaches the agent here, and it is named explicitly.
+          # changed nothing because the API key was never consulted.
           ANTHROPIC_BASE_URL: ${{ env.AGENT_BASE_URL }}
-          ANTHROPIC_AUTH_TOKEN: ${{ secrets.AI_BUILDER_API_KEY }}
+          AI_BUILDER_API_KEY: ${{ secrets.AI_BUILDER_API_KEY }}
+          ANTHROPIC_AUTH_TOKEN: ""
           ANTHROPIC_API_KEY: ""
           MAX_BUDGET_USD: ${{ vars.AI_BUILDER_MAX_USD || '2' }}
           DEBUG_OUTPUT: ${{ inputs.debug_output || false }}
         run: |
           set -euo pipefail
-          if [ -z "${ANTHROPIC_AUTH_TOKEN}" ]; then
+          if [ -z "${AI_BUILDER_API_KEY}" ]; then
             echo "::error::Missing AI_BUILDER_API_KEY secret"; exit 1
+          fi
+          if [ "${ANTHROPIC_BASE_URL%/}" = "https://api.anthropic.com" ]; then
+            export ANTHROPIC_API_KEY="$AI_BUILDER_API_KEY"
+          else
+            export ANTHROPIC_AUTH_TOKEN="$AI_BUILDER_API_KEY"
           fi
           # Pinned: this process holds the LLM key in its env, so a hijacked
           # release of the CLI would run holding that secret. Bump deliberately.
--- a/.github/workflows/ai-bug-fixer.yml
+++ b/.github/workflows/ai-bug-fixer.yml
@@ -283,8 +283,10 @@ jobs:
           # unavoidably in this process's env (the Read tool can see
           # /proc/self/environ), so: WebFetch/WebSearch are denied to close the
           # outbound-exfil channel; repo/API tokens are step-scoped OUT of this
-          # step; and later steps never execute agent-editable files (Finalize
-          # commits with --no-verify).
+          # step; later steps never execute agent-editable files (Finalize
+          # commits with --no-verify); and the guard rejects any edit under
+          # .github/ or .husky/ so agent-authored CI never reaches a
+          # `pull_request` run that holds secrets.
           # --max-budget-usd is a hard spend ceiling; unlike --max-turns it
           # bounds cost directly, which matters now billing is per-token rather
           # than a flat subscription fee.
@@ -312,15 +314,28 @@ jobs:
           fi
           exit "$AGENT_EXIT"
 
-      - name: Guard — no schema or migration edits
+      - name: Guard — no CI/hook edits; no schema edits (ai-buildable)
         id: guard
         continue-on-error: true
         run: |
           set -euo pipefail
-          # Features briefed via `ai-buildable` are told to STOP if they need a
-          # migration. That is an instruction, and a model can ignore an
-          # instruction — it cannot ignore a diff. Enforced only for the
-          # feature profile: an `ai-fixable` bug fix that legitimately needs a
+          # A model can ignore an instruction — it cannot ignore a diff.
+          #
+          # CI and hook paths are banned for BOTH profiles: they are the
+          # worker's own control plane. Finalize's `git add -A` commits, pushes
+          # and opens a PR from an in-repo branch, so `pull_request` workflows
+          # (test.yml, pr_agent.yml, …) would immediately EXECUTE an
+          # agent-authored workflow file with this repo's secrets, before any
+          # human review. Same for .husky/ hooks.
+          FORBIDDEN=$(git status --porcelain -- .github .husky)
+          if [ -n "$FORBIDDEN" ]; then
+            echo "::error::agent modified CI/hook files, which is never allowed"
+            echo "$FORBIDDEN"
+            exit 1
+          fi
+          # The schema/migration ban is `ai-buildable`-only. Features briefed
+          # via that label are told to STOP if they need a migration; this
+          # enforces it. An `ai-fixable` bug fix that legitimately needs a
           # schema change still goes through human review as before.
           #
           # Rationale: preview deploys run `next build` only, never
@@ -328,7 +343,7 @@ jobs:
           # database. The feature would look broken for reasons unrelated to
           # the agent's work.
           if [ "$TRIGGER_LABEL" != "ai-buildable" ]; then
-            echo "guard not applicable for '$TRIGGER_LABEL'"; exit 0
+            echo "guard (schema) not applicable for '$TRIGGER_LABEL'"; exit 0
           fi
           # --porcelain over `git diff` so NEW untracked migration directories
           # are caught too — a migration is almost always a new file, which a
@@ -415,7 +430,7 @@ jobs:
 
           if [ -n "$BAILED" ]; then fail "the agent flagged this as needing a human — ${BAILED}"; fi
           if [ "${{ steps.agent.outcome }}" != "success" ]; then abort "the coding agent step failed (log tail in the step above)"; fi
-          if [ "${{ steps.guard.outcome }}" != "success" ]; then fail "the agent changed the database schema, which ai-buildable forbids"; fi
+          if [ "${{ steps.guard.outcome }}" != "success" ]; then fail "the agent modified forbidden paths (CI/hook files, or schema/migrations under ai-buildable)"; fi
           if [ -z "$HAS_CHANGES" ]; then fail "the agent made no code changes"; fi
           if [ "${{ steps.validate.outcome }}" != "success" ]; then fail "validation (tsc/next lint) failed"; fi
 
--- a/.github/workflows/ai-bug-fixer.yml
+++ b/.github/workflows/ai-bug-fixer.yml
@@ -253,7 +253,7 @@ jobs:
           ANTHROPIC_AUTH_TOKEN: ${{ secrets.AI_BUILDER_API_KEY }}
           ANTHROPIC_API_KEY: ""
         run: |
-          set -uo pipefail
+          set -euo pipefail
           if [ -z "${ANTHROPIC_AUTH_TOKEN}" ]; then
             echo "::error::Missing AI_BUILDER_API_KEY secret"; exit 1
           fi
--- a/.github/workflows/ai-bug-fixer.yml
+++ b/.github/workflows/ai-bug-fixer.yml
@@ -189,9 +189,11 @@ jobs:
       # model id). Clearing the var does NOT do that: `||` here cannot tell unset
       # from empty, so both fall back to the OpenRouter default.
       AGENT_BASE_URL: ${{ vars.AI_BUILDER_BASE_URL || 'https://openrouter.ai/api' }}
-      EXPONENTIAL_TOKEN: ${{ secrets.EXPONENTIAL_TOKEN }}
-      # Prefer a PAT so the PR triggers CI; fall back to the default token.
-      GH_TOKEN: ${{ secrets.AI_BUG_FIXER_GH_TOKEN || github.token }}
+      # EXPONENTIAL_TOKEN and GH_TOKEN are deliberately NOT job-level env:
+      # job env is inherited by EVERY step, including the agent process (whose
+      # Read tool can read /proc/self/environ) and Validate's `npx next lint`
+      # (which executes agent-editable JS config). They are step-scoped to the
+      # steps that actually use them instead.
     steps:
       - uses: actions/checkout@v4
         with:
@@ -204,6 +206,10 @@ jobs:
           cache: npm
 
       - name: Install deps + CLI
+        # The only step that needs the raw token in env — `auth login` stores
+        # credentials for the CLI, so later steps call it with no token in env.
+        env:
+          EXPONENTIAL_TOKEN: ${{ secrets.EXPONENTIAL_TOKEN }}
         run: |
           npm ci
           npm i -g exponential-cli
@@ -271,10 +277,14 @@ jobs:
           # carries Bash(...) allow rules that merge into the agent's
           # permissions on top of these flags. A deny beats every allow source,
           # so this is what actually keeps the agent from shelling out — the
-          # brief forbids git, the workflow owns commit/push/PR, and an agent
-          # that cannot shell out cannot read the API key from its env.
-          # WebFetch/WebSearch are denied too: the brief needs neither, and a
-          # file-editing agent has no business making outbound requests.
+          # brief forbids git, and the workflow owns commit/push/PR.
+          #
+          # Containment is layered, not a single wall. The LLM key is
+          # unavoidably in this process's env (the Read tool can see
+          # /proc/self/environ), so: WebFetch/WebSearch are denied to close the
+          # outbound-exfil channel; repo/API tokens are step-scoped OUT of this
+          # step; and later steps never execute agent-editable files (Finalize
+          # commits with --no-verify).
           # --max-budget-usd is a hard spend ceiling; unlike --max-turns it
           # bounds cost directly, which matters now billing is per-token rather
           # than a flat subscription fee.
@@ -354,6 +364,9 @@ jobs:
           GIT_AUTHOR_EMAIL: ai-bot@bots.exponential.im
           GIT_COMMITTER_NAME: exponential-ai-bot
           GIT_COMMITTER_EMAIL: ai-bot@bots.exponential.im
+          # Prefer a PAT so the PR triggers CI; fall back to the default token.
+          # Step-scoped (see the job env comment): only this step runs `gh`.
+          GH_TOKEN: ${{ secrets.AI_BUG_FIXER_GH_TOKEN || github.token }}
         run: |
           set -euo pipefail
 
@@ -422,7 +435,11 @@ jobs:
           Autonomous change for Exponential ticket #${TICKET_NUMBER}.
           Requires human review before merge.
           EOF
-            git commit -F .git-commit-msg
+            # --no-verify: the workflow owns this commit. Hooks under .husky/
+            # are tracked, agent-editable files (installed by npm ci's
+            # `prepare`), so running them here would execute agent-authored
+            # code in a step that holds tokens.
+            git commit --no-verify -F .git-commit-msg
             rm -f .git-commit-msg
           fi
 
--- a/.github/workflows/ai-bug-fixer.yml
+++ b/.github/workflows/ai-bug-fixer.yml
@@ -138,6 +138,14 @@ jobs:
           exponential tickets list \
             --product "$EXPONENTIAL_PRODUCT" --workspace "$EXPONENTIAL_WORKSPACE" \
             --status READY_TO_PLAN --label security --json > security.json
+          # Retry bound: a ticket that failed a previous attempt carries
+          # `ai-attempted` (added on release, below). Excluding it here stops a
+          # deterministically-failing ticket from burning metered budget every
+          # hour forever AND (selection is oldest-first) starving the queue
+          # behind it. A human re-arms the ticket by removing the label.
+          exponential tickets list \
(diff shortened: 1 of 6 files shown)


scripts/ai-bug-fixer/select-candidate.mjs
```diff
@@ -53,7 +72,19 @@ function slugify(title) {
...
 }
 
-const candidates = readTickets(arg("candidates"));
+// Merge every --candidates file, stamping each ticket with the label whose
+// query produced it. First match wins on dedupe: a ticket carrying BOTH labels
+// is worked as `ai-fixable`, the stricter (narrow-fix) profile.
+const candidateFiles = args("candidates");
+const candidateLabels = args("candidate-label");
+/** @type {Map<string, any>} */
+const byId = new Map();
```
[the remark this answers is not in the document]
positonic: ✅ Addressed in [`5125f656`](https://github.com/positonic/exponential/pull/438/commits/5125f656) — mispaired `--candidates`/`--candidate-label` flags now exit 1 so the scan job goes red instead of silently selecting with an empty `triggerLabel`.

.github/workflows/ai-bug-fixer.yml
```diff
@@ -204,31 +237,79 @@ jobs:
...
+        env:
+          # ANTHROPIC_AUTH_TOKEN (not ANTHROPIC_API_KEY) is the header the SDK
+          # sends to a custom base URL. ANTHROPIC_API_KEY is blanked explicitly
+          # so a leftover Anthropic key in repo secrets cannot win the race and
+          # silently bill the wrong account.
+          ANTHROPIC_BASE_URL: ${{ env.AGENT_BASE_URL }}
+          ANTHROPIC_AUTH_TOKEN: ${{ secrets.AI_BUILDER_API_KEY }}
+          ANTHROPIC_API_KEY: ""
+        run: |
+          set -uo pipefail
```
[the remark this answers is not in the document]
positonic: ✅ Addressed in [`af5aceb3`](https://github.com/positonic/exponential/pull/438/commits/af5aceb3) — the agent step now starts with `set -euo pipefail`; the existing `set +e`/`set -e` toggle around the `claude` invocation still captures its exit code.

.github/workflows/ai-bug-fixer.yml
```diff
@@ -204,31 +237,86 @@ jobs:
...
+          # --max-budget-usd is a hard spend ceiling; unlike --max-turns it
+          # bounds cost directly, which matters now billing is per-token rather
+          # than a flat subscription fee.
+          set +e
+          claude -p "Read .ai-bug-fixer/prompt.md and implement what it describes. Edit the working tree only — do NOT run git, commit, push, or open a PR." \
+            --model "$MODEL" \
+            --max-turns 30 \
+            --max-budget-usd "${{ vars.AI_BUILDER_MAX_USD || '2' }}" \
+            --permission-mode acceptEdits \
+            --allowedTools "Read,Write,Edit,Glob,Grep" \
```
positonic: > ✅ Addressed in [`e756aa14`](https://github.com/positonic/exponential/pull/438/commits/e756aa14) — tokens step-scoped out of the agent/lint env; Finalize commits with --no-verify; containment comment narrowed to the layered claim (WebFetch/WebSearch were already denied in 80983cf8)

---

**[High] The "no Bash" containment is porous — agent-written files execute in later steps that hold secrets**

Withholding Bash stops direct shell-out, but the agent's `Write`/`Edit` output is *executed* by later steps in this job, which carry `GH_TOKEN` (possibly a PAT) and `EXPONENTIAL_TOKEN` in job-level env: (1) `npm ci` ran husky's `prepare`, so Finalize's `git commit` executes `.husky/pre-commit` — a tracked file the agent can edit; (2) Validate's `npx next lint` loads `next.config.js` and the eslint config, both agent-editable JS; (3) the `Read` tool can read `/proc/self/environ` (the agent process inherits job env, including `EXPONENTIAL_TOKEN`/`GH_TOKEN`) and `Write` the values into a file that Finalize commits and pushes to a public branch.

Suggested fix (three small changes):
```yaml
# 1. Finalize: the workflow owns the commit — don't run agent-editable hooks
git commit --no-verify -F .git-commit-msg

# 2. Move EXPONENTIAL_TOKEN / GH_TOKEN out of the fix job's `env:` block and
#    into only the steps that need them (Install+auth CLI, Claim, Finalize),
#    so neither the agent process nor Validate has them in env.

# 3. (belt-and-braces) add --disallowedTools "WebFetch,WebSearch" — with a key
#    in env and network egress, fetch is an exfil channel too.
```

<details>

In `.github/workflows/ai-bug-fixer.yml`, the "Run coding agent" step (~line 262) withholds Bash from the Claude Code CLI on the theory that the agent then can't exfiltrate credentials or run code. But later steps in the same job execute files the agent can write: Finalize's `git commit` runs `.husky/pre-commit` (husky is installed by `npm ci`'s `prepare` script), and Validate's `npx next lint` loads `next.config.js`/eslint config. Job-level `env:` also puts `EXPONENTIAL_TOKEN` and `GH_TOKEN` into every step's environment, including the agent's own process (readable via the Read tool at `/proc/self/environ`).

Make these changes: (1) in the Finalize step, change `git commit -F .git-commit-msg` to `git commit --no-verify -F .git-commit-msg`; (2) remove `EXPONENTIAL_TOKEN` and `GH_TOKEN` from the `fix` job's top-level `env:` block and add them as step-level `env:` only on the steps that use them ("Install deps + CLI", "Claim ticket", "Finalize"; the checkout `token:` input already has its own expression); (3) add `--disallowedTools "WebFetch,WebSearch"` to the `claude` invocation. Constraints: don't change the step ordering or the claim/release logic; keep the workflow's existing comments accurate — update the "an agent that cannot shell out cannot exfiltrate it" comment to reflect the narrowed claim. Verify the workflow still parses (`actionlint` or a YAML parse) after the env restructuring.

</details>

.github/workflows/ai-bug-fixer.yml
```diff
@@ -204,31 +237,86 @@ jobs:
...
+          # `prisma migrate deploy`, so a schema change never reaches a preview
+          # database. The feature would look broken for reasons unrelated to
+          # the agent's work.
+          if [ "$TRIGGER_LABEL" != "ai-buildable" ]; then
+            echo "guard not applicable for '$TRIGGER_LABEL'"; exit 0
+          fi
+          # --porcelain over `git diff` so NEW untracked migration directories
+          # are caught too — a migration is almost always a new file, which a
+          # plain diff against HEAD would not see.
+          CHANGED=$(git status --porcelain -- prisma/schema.prisma prisma/migrations)
```
positonic, line 377: > ✅ Addressed in [`7d3c2370`](https://github.com/positonic/exponential/pull/438/commits/7d3c2370) — .github/.husky edits now fail the guard for BOTH profiles, before the label early-exit; prisma check stays ai-buildable-only; release message and docs updated

---

**[High] Guard bans only prisma paths — workflow-file edits are pushed and executed by `pull_request` CI with repo secrets**

The guard's pathspec covers only `prisma/`. If the agent writes to `.github/workflows/` (or `.husky/`), Finalize's `git add -A` commits it, pushes it, and opens a PR — and because the branch is in-repo (not a fork), `pull_request`-triggered workflows (`test.yml`, `pr_agent.yml`, …) immediately run the *agent-authored* workflow definition with this repo's secrets available. No human review happens before that execution. Also note the guard is skipped entirely for `ai-fixable`.

Suggested fix — ban CI/hook paths unconditionally, keep the prisma ban profile-scoped:
```bash
# Applies to BOTH profiles: these paths are the worker's own control plane.
FORBIDDEN=$(git status --porcelain -- .github .husky)
if [ -n "$FORBIDDEN" ]; then
  echo "::error::agent modified CI/hook files, which is never allowed"
  echo "$FORBIDDEN"; exit 1
fi
if [ "$TRIGGER_LABEL" != "ai-buildable" ]; then
  echo "guard (schema) not applicable for '$TRIGGER_LABEL'"; exit 0
fi
CHANGED=$(git status --porcelain -- prisma/schema.prisma prisma/migrations)
```

<details>

In `.github/workflows/ai-bug-fixer.yml`, the "Guard — no schema or migration edits" step (~line 292–319) checks `git status --porcelain -- prisma/schema.prisma prisma/migrations`, and only when `TRIGGER_LABEL` is `ai-buildable`. Extend it so that edits under `.github/` and `.husky/` fail the guard for **both** profiles (move that check above the early `exit 0` for non-`ai-buildable` labels), while the prisma check stays `ai-buildable`-only. Rationale: the branch is pushed and a PR is opened automatically; same-repo `pull_request` workflows then execute any agent-modified workflow file with repo secrets before any human looks at it.

Constraints: keep `set -euo pipefail` semantics; keep the step's `continue-on-error: true`; update the corresponding `fail` message in the Finalize step (~line 389) so a guard failure no longer claims it was necessarily a schema change — e.g. "the agent modified forbidden paths (schema/migrations or CI files)". Also update the guard description in `docs/agents/ai-bug-fixer.md` (the "two profiles" table and the "Guard" bullet) to match.

</details>

.github/workflows/ai-bug-fixer.yml
```diff
@@ -153,7 +178,15 @@ jobs:
...
-      MODEL: ${{ inputs.model || vars.AI_BUG_FIXER_MODEL || 'claude-haiku-4-5' }}
+      TRIGGER_LABEL: ${{ needs.scan.outputs.trigger_label }}
+      # Deliberately a NEW var name rather than reusing AI_BUG_FIXER_MODEL: that
+      # one named an Anthropic model, and a stale value left in repo settings
+      # would be sent to OpenRouter, which does not know it.
+      MODEL: ${{ inputs.model || vars.AI_BUILDER_MODEL || 'z-ai/glm-5.2' }}
+      # OpenRouter's Anthropic-compatible endpoint. Note `/api`, NOT `/api/v1` —
+      # the Anthropic SDK appends its own path. Unset this var to talk to
+      # Anthropic directly instead (then MODEL must be an Anthropic model id).
+      AGENT_BASE_URL: ${{ vars.AI_BUILDER_BASE_URL || 'https://openrouter.ai/api' }}
```
positonic, line 200: > ✅ Addressed in [`3dc5bb47`](https://github.com/positonic/exponential/pull/438/commits/3dc5bb47) — the step now exports the key as ANTHROPIC_API_KEY (x-api-key) for api.anthropic.com and ANTHROPIC_AUTH_TOKEN (Bearer) otherwise — one credential either way; the 'clear the var' wording was already fixed in 80983cf8

---

**[Medium] The documented "clear `AI_BUILDER_BASE_URL` to talk to Anthropic directly" escape hatch cannot work**

In GitHub expressions an empty/unset var is falsy, so `vars.AI_BUILDER_BASE_URL || 'https://openrouter.ai/api'` yields the OpenRouter default precisely when the var is cleared — the state the comment (and `docs/agents/ai-bug-fixer.md`, and the ADR amendment) says means "talk to Anthropic directly". There is no value of this var that produces an unset `ANTHROPIC_BASE_URL`. Separately, even with the var set to `https://api.anthropic.com`, auth is passed as `ANTHROPIC_AUTH_TOKEN` (sent as `Authorization: Bearer`), which is not how an `sk-ant-api…` key normally authenticates (`x-api-key`, sourced from `ANTHROPIC_API_KEY` — which this step deliberately blanks).

Suggested fix: document the real fallback (set the var explicitly to `https://api.anthropic.com`), and verify/handle the auth-header question for that path. Update the three places that document "clear/unset the var": this comment, `docs/agents/ai-bug-fixer.md`, and the ADR-0029 amendment.

<details>

In `.github/workflows/ai-bug-fixer.yml` (~lines 186–189), `AGENT_BASE_URL: ${{ vars.AI_BUILDER_BASE_URL || 'https://openrouter.ai/api' }}` is documented (here, in `docs/agents/ai-bug-fixer.md`, and in `docs/adr/0029-ai-bug-fixer-workflow.md`) as "unset/clear the var to talk to Anthropic directly". That's impossible: an empty GitHub Actions variable is falsy, so clearing it selects the OpenRouter default. Fix the documentation in all three files to say the fallback is setting `AI_BUILDER_BASE_URL=https://api.anthropic.com` explicitly. Then check whether the direct-Anthropic path authenticates at all: the agent step passes the key only as `ANTHROPIC_AUTH_TOKEN` (Bearer header) and blanks `ANTHROPIC_API_KEY`; an Anthropic console API key normally authenticates via the `x-api-key` header, which the SDK sends from `ANTHROPIC_API_KEY`. If Bearer doesn't work against api.anthropic.com, either document that the fallback requires editing the step, or make the step conditionally export the key under the right variable based on the base URL. Constraints: keep exactly one credential secret (`AI_BUILDER_API_KEY`); do not reintroduce the two-credential precedence trap the comments in this step warn about.

</details>

.github/workflows/ai-bug-fixer.yml
```diff
@@ -204,31 +237,86 @@ jobs:
...
+          # commit/push/PR, and this process holds an API key in its env — in a
+          # public repo, an agent that cannot shell out cannot exfiltrate it.
+          # --max-budget-usd is a hard spend ceiling; unlike --max-turns it
+          # bounds cost directly, which matters now billing is per-token rather
+          # than a flat subscription fee.
+          set +e
+          claude -p "Read .ai-bug-fixer/prompt.md and implement what it describes. Edit the working tree only — do NOT run git, commit, push, or open a PR." \
+            --model "$MODEL" \
+            --max-turns 30 \
+            --max-budget-usd "${{ vars.AI_BUILDER_MAX_USD || '2' }}" \
```
positonic: > ✅ Addressed in [`58e88ca4`](https://github.com/positonic/exponential/pull/438/commits/58e88ca4) — workflow comment, docs and ADR no longer present the flag as a hard ceiling; key-level OpenRouter spend limit documented as the enforced backstop; empirical cost check folded into the planned first smoke run (no key available yet)

---

**[Medium] `--max-budget-usd` may be a no-op for a model Claude Code can't price**

The budget ceiling works by Claude Code computing cost from its internal per-model pricing table. `z-ai/glm-5.2` served through a proxy is not a model ID it knows; unknown models routinely report `$0.00` total cost, in which case the ceiling never binds and `--max-turns 30` is the only real bound. Since the ADR amendment explicitly names this flag as the compensating control for losing flat-fee billing, this belongs in the already-planned end-to-end smoke test: run once, then confirm the reported cost is non-zero. If it's zero, say so in the docs and lean on `--max-turns` plus an OpenRouter key-level spend limit instead.

<details>

In `.github/workflows/ai-bug-fixer.yml` (~line 273), the agent invocation passes `--max-budget-usd` as a hard per-run spend ceiling, with the model served via OpenRouter's Anthropic-compatible endpoint (`ANTHROPIC_BASE_URL=https://openrouter.ai/api`, model `z-ai/glm-5.2`). Claude Code computes spend from its internal pricing table keyed by model ID; verify whether a non-Anthropic model ID through a custom base URL yields a non-zero computed cost. Do this empirically: run `claude -p "say hi" --model z-ai/glm-5.2 --max-budget-usd 0.01 --output-format json` against the OpenRouter base URL and check `total_cost_usd` in the result, and/or test that an absurdly low budget (e.g. `0.000001`) actually terminates a longer run. If the ceiling doesn't bind, update `docs/agents/ai-bug-fixer.md` ("Cost controls") and the ADR-0029 amendment to stop presenting it as a hard ceiling, and recommend an OpenRouter key-level spend limit as the real backstop. Do not remove the flag — it's harmless if inert and correct if pricing metadata arrives later.

</details>

.github/workflows/ai-bug-fixer.yml
```diff
@@ -298,48 +386,47 @@ jobs:
 
           if [ -n "$BAILED" ]; then fail "the agent flagged this as needing a human — ${BAILED}"; fi
           if [ "${{ steps.agent.outcome }}" != "success" ]; then fail "the coding agent step failed"; fi
+          if [ "${{ steps.guard.outcome }}" != "success" ]; then fail "the agent changed the database schema, which ai-buildable forbids"; fi
```
positonic: > ✅ Addressed in [`462b222d`](https://github.com/positonic/exponential/pull/438/commits/462b222d) — release now adds ai-attempted (best-effort); scan excludes it via a second --exclude file (flag now repeatable); label created in syntrofi; re-arm by removing the label, documented

---

**[Medium] A deterministically-failing ticket now burns budget every hour forever, and blocks the queue**

`fail` releases the ticket back to `READY_TO_PLAN`, and the scan picks the **oldest** eligible ticket — i.e. the same one — next hour. Under flat-fee billing an infinite retry loop was free; this PR makes it metered: up to `AI_BUILDER_MAX_USD × 24`/day, indefinitely, with no attempt counter. The new guard makes deterministic loops *more* likely (a ticket whose feature genuinely needs a migration will trip it every single run), and FIFO means the looping ticket also starves every newer ticket behind it.

Suggested fix: bound attempts. Cheapest version — on `fail`, add an `ai-attempted` label; in the scan, exclude tickets already carrying it (one more `tickets list` + exclude file, same pattern as `security`). A human re-arms a ticket by removing the label. Alternatively count the worker's own 🤖 release comments in `select-candidate.mjs` and skip after N.

<details>

In `.github/workflows/ai-bug-fixer.yml`, the Finalize step's `fail()` path releases a ticket back to `READY_TO_PLAN`, and `scripts/ai-bug-fixer/select-candidate.mjs` picks the oldest eligible ticket — so a ticket that fails deterministically (guard trip, un-implementable spec, validation failure) is re-attempted every hourly run forever, now at per-token cost, while blocking newer tickets (FIFO). Add a retry bound: in the Finalize `release_ticket` function, also `--add-label ai-attempted`; in the scan job, add a fourth `exponential tickets list … --label ai-attempted --json > attempted.json` query and pass it to `select-candidate.mjs` as a second `--exclude` file (make `--exclude` repeatable the same way `--candidates` already is). Document in `docs/agents/ai-bug-fixer.md` that removing `ai-attempted` re-arms a ticket. Constraints: the `ai-attempted` label must exist in the workspace before merge (the CLI errors on unknown label slugs — same load-bearing detail as `ai-buildable`); label operations on release must stay best-effort (`|| true`) so they can't strand a ticket; keep exit code 0 semantics in select-candidate.mjs.

</details>

docs/agents/ai-bug-fixer.md
```diff
@@ -1,46 +1,79 @@
...
 merge → ticket `DONE`) is handled by the existing webhook in
 [ADR-0021](../adr/0021-pr-merge-promotes-ticket-via-app-webhook.md).
 
-## How a bug becomes eligible
+## The two profiles
+
+| | `ai-fixable` | `ai-buildable` |
+| --- | --- | --- |
+| For | bugs | features |
+| Brief framing | smallest change that fixes it | build what the ticket describes |
```
positonic: > ✅ Addressed in [`2587c9c8`](https://github.com/positonic/exponential/pull/438/commits/2587c9c8) — render-prompt.mjs takes --label and appends the schema-ban/bail rule for ai-buildable regardless of ticket type; adds render-prompt.test.ts; docs now attribute framing to type, hard rules to label

---

**[Low] "Brief framing" is keyed on ticket `type`, not the trigger label — this table implies otherwise**

`render-prompt.mjs` chooses framing from `ticket.type` (`isBug`), and never sees the trigger label; this table attributes framing to the labels. The gap is concrete for a `type = BUG` ticket labelled `ai-buildable`: it gets the narrow-fix brief, whose bail rule does **not** mention migrations — yet the migration guard applies. If that fix needs a schema change, the agent won't know to bail, the guard trips, and the ticket loops hourly. (Inverse case: a feature-typed `ai-fixable` ticket is told to stop on migrations but has no guard.)

Suggested fix: pass `TRIGGER_LABEL` into `render-prompt.mjs` (`--label` flag) and add one line to the brief when the label is `ai-buildable`: "This task may NOT touch `prisma/schema.prisma` or `prisma/migrations` — if it requires that, STOP and write `.ai-bug-fixer/needs-human.txt`." Then this table's framing row is honest (or reword it to say framing follows ticket type).

<details>

In this repo, `scripts/ai-bug-fixer/render-prompt.mjs` builds the agent brief and chooses its framing from `ticket.type` (the `isBug` constant), while the workflow `.github/workflows/ai-bug-fixer.yml` selects the profile (and the schema/migration guard) from the trigger label in `TRIGGER_LABEL`. Wire the label into the brief: add a `--label` CLI arg to `render-prompt.mjs`, pass `"$TRIGGER_LABEL"` from the "Render fix brief" step, and when the label is `ai-buildable` append an explicit rule to the brief: schema/migration edits are forbidden and the agent must STOP and write `.ai-bug-fixer/needs-human.txt` if the ticket requires them. Keep `buildBrief` pure and exported (it has tests in render-prompt.test.ts — extend them for the new parameter, defaulting to existing behavior when the label is absent). Also correct the "Brief framing" row in `docs/agents/ai-bug-fixer.md`'s profile table to reflect what actually drives framing.

</details>

.github/workflows/ai-bug-fixer.yml
```diff
@@ -204,31 +237,86 @@ jobs:
...
+          # credential reaches the agent here, and it is named explicitly.
+          ANTHROPIC_BASE_URL: ${{ env.AGENT_BASE_URL }}
+          ANTHROPIC_AUTH_TOKEN: ${{ secrets.AI_BUILDER_API_KEY }}
+          ANTHROPIC_API_KEY: ""
+        run: |
+          set -euo pipefail
+          if [ -z "${ANTHROPIC_AUTH_TOKEN}" ]; then
+            echo "::error::Missing AI_BUILDER_API_KEY secret"; exit 1
+          fi
+          npm i -g @anthropic-ai/claude-code
```
positonic: > ✅ Addressed in [`80983cf8`](https://github.com/positonic/exponential/pull/438/commits/80983cf8) — CLI pinned to @anthropic-ai/claude-code@2.1.186 with a deliberate-bump comment (landed in the previous commit)

---

**[Low] Pin the Claude Code CLI version**

The PR notes `--max-turns` is already undocumented in 2.1.186 and was verified by trial — that's exactly the kind of flag that vanishes in an unpinned `npm i -g`. A CLI release that drops or renames any of the five flags used here fails every hourly run (each one claiming and releasing a ticket) until someone reads the log tail. An unpinned binary that runs with a live API key in a public repo's runner is also an avoidable supply-chain surface.

Suggested fix:
```suggestion
          npm i -g @anthropic-ai/claude-code@2.1.186
```
Bump deliberately, re-verifying the flag set each time.

<details>

In `.github/workflows/ai-bug-fixer.yml` (~line 260), the "Run coding agent" step installs the CLI with `npm i -g @anthropic-ai/claude-code` (no version). Pin it to the version the PR was verified against (`@2.1.186`), so an upstream release that changes flags (`--max-turns` is already undocumented; `--max-budget-usd` is new) can't break the hourly worker unannounced. Optionally add a comment noting the bump procedure: change the pin, run once via workflow_dispatch, confirm a PR is produced. Don't pin `exponential-cli` in the same change unless asked — it's a different trade-off.

</details>

PR #459, opened 2026-08-03
feat: local wiki + librarian (query + write-back)

positonic:
## What

The foray's probe feature: a device-local, git-backed markdown wiki (the Karpathy LLM-wiki pattern) maintained by a librarian agent, reachable from ManyChat **only inside the Tauri shell**. Query + write-back; ingest and lint are V3.

A wiki turn does not go through this backend. The webview streams straight to Mastra, the wiki tools execute on the user's machine over Tauri IPC, and the only Vercel call in the whole flow is the one-time `mastra.mintAgentToken`.

Companion PR: positonic/mastra#57 (the `localWikiAgent` itself).

## The parts worth reviewing

**The jail.** These commands are reachable from a remote origin, so a path bug here is a read/write primitive over the user's disk handed to whatever the model was talked into emitting. `resolve` checks twice on purpose: it rejects `..`, absolute paths and prefixes before touching the disk (which catches traversal to a path that doesn't exist yet — the write case), then canonicalizes the deepest existing ancestor against the root (the only thing that catches a symlink already sitting in the wiki). Either check alone has a hole.

**No setter for the wiki root.** The ticket allowed a bridge command for this; I didn't add one. A jail whose walls the caller can move isn't a jail — a page-callable setter would turn "read inside the wiki" into "read anywhere" for any script on that origin. The root is settled once at startup (env override → stored value → default), persisted, and exposed read-only.

**One commit per turn, not per write.** Filing a page, linking it from `index.md` and appending to `log.md` are one change. Per-write commits would let you revert the page and leave the index pointing at a file that no longer exists. The transport commits once at turn end, only if a write actually *succeeded* — a write refused by the jail changed nothing, and committing for it would produce an empty commit implying otherwise.

**`schema.md` is the contract, and the user owns it.** It carries naming, `[[wikilinks]]`, index/log discipline and multi-agent etiquette. Seeded once, never overwritten, and the agent's instructions say that when it and the file disagree, the file wins. That's what keeps the wiki operable by Claude Code or an MCP-connected agent later.

**Two constraints inherited from the spike.** Tool schemas are hand-written JSON Schema, never zod (client-js mangles this app's zod and the provider kills the turn); and the picker entry is injected client-side rather than added to `getMastraAgents`' allow-list, which feeds every surface and would offer the librarian on web and Electron where every turn fails on the first tool call.

## Verification

`cargo test` — 47 tests, including the jail (traversal, absolute paths, symlink escape on both read and write paths), init idempotence against a user-edited `schema.md`, commit-per-turn, search bounds, and that moving the root moves the walls with it.

`streamLocalWikiChat` — 17 unit tests on the same seam as `streamProtocol.test.ts`: synthetic client-js chunks in, `ChatStreamUpdate` out. Tool lifecycle, both error-frame shapes, unrecognised chunks ignored, and every branch of the commit rule.

Against a release build loading production, with the root pointed at a scratch folder: init created it there and left the default untouched, three writes in one turn produced exactly one commit containing all of them, a following read-only turn committed nothing, search found the page, and `../escape.md` was refused.

**Not verified end-to-end:** a live librarian turn, because `localWikiAgent` isn't on the deployed Mastra server until positonic/mastra#57 lands and is deployed (`railway up`). Both halves are proven separately — the clientTools round-trip against production in #458, and the wiki commands from the production page here.

## Acceptance criteria

- [x] Nine V2 requirement rows (seeded git wiki; Tauri-only picker entry; direct webview↔Mastra streaming with local tool execution; path jail; commit per writing turn; write-back with index/log conventions; brain-agnostic format + contract; text-only search; no Vercel backend in the turn data path)
- [x] `cargo test` covers the jail and commit-per-turn; transport unit tests follow the `streamProtocol.test.ts` pattern
- [x] Web and Electron show no trace of the local-wiki entry — the picker entry is gated on the Tauri bridge and the agent is absent from the server-side allow-list

---

Ships: neon.puffin

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Adds device-local git-backed markdown wiki with Tauri IPC commands (init, read, write, search, commit)
  - Jailed filesystem access with double-check path traversal prevention
  - Idempotent `wiki_init` seeds `schema.md`, `index.md`, `log.md` on first run
  - One commit per turn (not per write) preserves coherent git history

- Introduces `streamLocalWikiChat` transport: browser streams directly to Mastra, wiki content never passes through Vercel

- Adds `mintAgentToken` tRPC procedure replacing throwaway dev-only JWT route

- Injects "Local wiki" librarian agent into chat picker only inside Tauri shell

___

### Diagram Walkthrough

```mermaid
flowchart LR
  ZoeDrawer["ZoeDrawer\n(agent picker)"]
  isLocalWikiAvailable["isLocalWikiAvailable()"]
  ManyChat["ManyChat\n(chat UI)"]
  mintAgentToken["mastra.mintAgentToken\n(tRPC)"]
  streamLocalWikiChat["streamLocalWikiChat\n(browser → Mastra)"]
  wikiBridge["WikiBridge\n(Tauri IPC)"]
  tauriWiki["src-tauri/wiki.rs\n(jailed FS + git)"]
  mastraServer["Mastra Server\n(localWikiAgent)"]

  ZoeDrawer -- "checks" --> isLocalWikiAvailable
  ZoeDrawer -- "shows librarian entry if available" --> ManyChat
  ManyChat -- "wiki turn only" --> mintAgentToken
  mintAgentToken -- "returns JWT + mastraUrl" --> ManyChat
  ManyChat -- "stream turn + clientTools" --> streamLocalWikiChat
  streamLocalWikiChat -- "tool calls" --> wikiBridge
  wikiBridge -- "IPC invoke" --> tauriWiki
  streamLocalWikiChat -- "stream" --> mastraServer
  tauriWiki -- "commitTurn after stream" --> tauriWiki
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>wiki.rs</strong><dd><code>New Rust module: jailed wiki FS commands with git history</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-c76367aec524f64c600801f6cccf9d93a2e5200c880815f90f56cf232b8fc8d7">+951/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>localWiki.ts</strong><dd><code>Wiki bridge interface, client tools, and agent constants</code>&nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-e26d7a1d853f45ef05efd3a482d3ae30793fe02feea92a6e87e731d169def149">+241/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>streamLocalWikiChat.ts</strong><dd><code>Browser-to-Mastra wiki chat transport with tool lifecycle</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-41a3f71390e43463cc1acb20823f082f67a752174f2bc4da45eff423ea1bc651">+258/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>ManyChat.tsx</strong><dd><code>Route wiki turns through local streamer bypassing backend</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-6f0e68ce865b0629e85c68dbd985a102db9f321cea87620c205f1863eece6ed4">+65/-19</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>ZoeDrawer.tsx</strong><dd><code>Inject local wiki librarian into agent picker for Tauri only</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-eb251afe64bce320035ae982586d6659cc574e66584644d41a0e864e05485aed">+39/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>mastra.ts</strong><dd><code>Add mintAgentToken protected tRPC mutation for browser JWT</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-afd76ed5d04839b106766464d541d99c9b1d0e8b28477ea826c7dcf52db960bb">+45/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ClientToolsSpike.tsx</strong><dd><code>Replace dev route token fetch with real tRPC procedure</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-13b3580903bf463445f5f6c20b8ac3b471b23a510b590d2a5ce4105899523e76">+11/-11</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>streamLocalWikiChat.test.ts</strong><dd><code>Comprehensive tests for wiki stream adapter and commit logic</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-f9f4008b42b7e3e17114d74eba0e4ac534727f4894092578ebd69e67fb3e139d">+309/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>test.yml</strong><dd><code>Add conditional Tauri Rust test job to CI pipeline</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88">+37/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>5 files</summary><table>
<tr>
  <td><strong>lib.rs</strong><dd><code>Register wiki module, commands, store plugin, and root init</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-eabcebd0ae5c9b77a5247e1bdbdb88b1869fc0c932d0ef0cf2ecf9ee5221be7c">+53/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>build.rs</strong><dd><code>Declare wiki commands for Tauri ACL autogeneration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-7d1ef5b0b986fee369d8f5fa4361f21fa3075601e5416738ba4f4d1ea5dfb2ed">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>remote.json</strong><dd><code>Grant wiki command permissions to remote origins</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-ccb4581457f281d685bf28be1cdf9c3dd911459508f59b4aecc48ebea72c5e44">+8/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>wiki_commit_turn.toml</strong><dd><code>Autogenerated ACL permissions for wiki_commit_turn command</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-18ca4ab25d03ecb2fabccf90d920fb28491afc8e4b0c071d5a2d487acad70f6f">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>wiki_get_root.toml</strong><dd><code>Autogenerated ACL permissions for wiki_get_root command</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-aa86c0af002a305760df90e044bc10b84ebc244e113919dccf88680732f9775e">+11/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Dependencies</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>Cargo.toml</strong><dd><code>Add tauri-plugin-store dependency for wiki root persistence</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-fe4c6a80a21e81339e2e0faeb9134d0f9636c168987df13c8abf1927a1caee39">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>schema.md</strong><dd><code>Seed wiki contract file for librarian conventions</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-7a1da3674d9ca9f9d3d8834e8b255ceec3278212f7c1cfe5901bed0f611d59b1">+60/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>index.md</strong><dd><code>Seed wiki index page template</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-520136298d408a3d282fd5e664da57ec24681be74ee9779eb9e74c96c74c0832">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>log.md</strong><dd><code>Seed wiki change log template</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-d802cbc78877e0f06d98b1a513fc02001d85d90356ad54517bdceee31e89bd2e">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>README.md</strong><dd><code>Document local wiki, security notes, and root configuration</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-f18b5d0480de73d6594622323ea9d3180f720e932124e6c58e03662d544aa4aa">+46/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Miscellaneous</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>route.ts</strong><dd><code>Remove throwaway dev-only JWT minting route</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-4449de74918b52c846eb30aee6cb23c04494c528ea7fcc97c5fbe76b9d66f4b7">+0/-40</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Additional files</strong></td><td><details><summary>5 files</summary><table>
<tr>
  <td><strong>wiki_init.toml</strong></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-922c4eb4159d8594c4c04235cd2c7a3a26ce02b4e4c7e4fbc0ec0797f72b4a4d">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>wiki_list_pages.toml</strong></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-17ba5a98ead7206397c9a88306ee450918e1892ee597ea7a65efac06b803ac2e">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>wiki_read_page.toml</strong></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-1ced4869ffbd8433079423dabeedec7c6ad3cc3f4c973684928f6bf6a537a954">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>wiki_search.toml</strong></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-4cff69f7284e7d0b70084ac06dace2513b293742a27ceeb80ecf6a536e2a23c6">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>wiki_write_page.toml</strong></td>
  <td><a href="https://github.com/positonic/exponential/pull/459/files#diff-27ff76b537a0685e8c1aa34e112b6f0c20294b88e75acd5537b38a7f01e5c63f">+11/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr></tr></tbody></table>

</details>

___


src-tauri/src/wiki.rs
```diff
@@ -0,0 +1,937 @@
...
+pub fn resolve(root: &Path, rel: &str) -> WikiResult<PathBuf> {
+    if rel.trim().is_empty() {
+        return Err(WikiError::BadPath);
+    }
+
+    let rel_path = Path::new(rel);
+    for component in rel_path.components() {
+        match component {
+            Component::Normal(_) => {}
+            Component::CurDir => {}
```
positonic, line 154: **[Critical] The jail lets the model write into `.git` — `core.fsmonitor` turns that into command execution**

`resolve` rejects `..`, absolute paths and prefixes, but `Component::Normal(".git")` sails through. A model talked into emitting `wiki_write_page(".git/config", "[core]\n\tfsmonitor = <command>")` plants config that git executes via the shell on the next index refresh — which `commit_turn`'s `git add -A` performs at the end of that very turn. That is arbitrary command execution on the user's machine, reachable purely from model output (e.g. prompt injection in a wiki page the librarian reads). The read side has the same gap: `wiki_read_page(".git/config")` leaks remote URLs, which can embed credentials.

Suggested fix — refuse dot-leading components, matching `collect_pages`' "dot = not content" rule (the librarian can never see these files anyway):

```suggestion
            Component::Normal(name) => {
                // Dotfiles are the wiki's machinery, not its content — and a
                // writable `.git/config` is a code-execution primitive
                // (`core.fsmonitor` runs on `commit_turn`'s `git add`).
                if name.to_string_lossy().starts_with('.') {
                    return Err(WikiError::BadPath);
                }
            }
            Component::CurDir => {}
```

Optional defense-in-depth: pass `-c core.fsmonitor= -c core.hooksPath=` on the `git` invocations in `git()` so a `.git/config` that arrives some other way (a cloned wiki) is inert too.

<details>

In `src-tauri/src/wiki.rs`, the `resolve` function's component loop (around lines 143–153) only rejects `ParentDir`, `RootDir`, and `Prefix` components. `Component::Normal(".git")` passes, so `wiki_write_page(".git/config", ...)` can plant `core.fsmonitor = <command>`, which git executes via the shell when `commit_turn` runs `git add -A` — arbitrary command execution from model-emitted tool arguments. `wiki_read_page(".git/config")` also leaks remote URLs.

Fix: in the `Component::Normal` arm, reject any component whose name starts with `.` (return `Err(WikiError::BadPath)`). This is consistent with `collect_pages`, which already treats dot-entries as non-content.

Constraints:
- Don't change `resolve`'s signature or the other match arms.
- Keep all existing tests green (none write dot-paths through `resolve`).
- Add regression tests: `resolve(root, ".git/config")`, `resolve(root, "notes/.git/config")`, and `resolve(root, ".hidden.md")` must all be refused, on both the read and write paths (`read_page`/`write_page`).
- Optionally harden the `git()` helper by prepending `-c core.fsmonitor= -c core.hooksPath=` to every invocation, but the path rejection is the primary fix.

</details>

src-tauri/src/wiki.rs
```diff
@@ -0,0 +1,937 @@
...
+    }
+
+    let canonical_root = root.canonicalize().map_err(|_| WikiError::NotFound)?;
+    let joined = canonical_root.join(rel_path);
+
+    // Walk up to the nearest existing ancestor: a write to a new page has no
+    // canonical path of its own yet, but its parent directory does.
+    let mut probe = joined.as_path();
+    let existing = loop {
+        if probe.exists() {
```
positonic, line 173: **[High] A dangling symlink slips past the ancestor walk — `fs::write` then creates a file outside the jail**

`Path::exists()` *follows* symlinks and returns `false` for a broken one. So for a symlink `note.md -> ~/.ssh/authorized_keys` sitting in the wiki **whose target does not exist yet**, the walk skips right past it up to the root, the root canonicalizes fine, and `resolve` returns the joined path — which `fs::write` then follows, creating the file *outside* the wiki. The existing symlink test only covers link targets that exist (where `exists()` is `true` and canonicalization catches the escape). A pre-placed symlink is required, but a cloned or shared wiki is exactly the multi-agent future `schema.md` describes, and git happily checks out symlinks.

Suggested fix — use lstat, which treats a dangling link as existing, so canonicalization runs on it and fails into `OutsideWiki`:

```suggestion
        if probe.symlink_metadata().is_ok() {
```

<details>

In `src-tauri/src/wiki.rs`, `resolve`'s nearest-existing-ancestor walk (around line 160) uses `probe.exists()`, which follows symlinks and returns `false` for a dangling symlink. Consequence: a symlink inside the wiki pointing at a *non-existent* outside path (e.g. `note.md -> ~/.ssh/authorized_keys` before that file exists) is skipped by the walk, `resolve` succeeds, and `write_page`'s `fs::write` follows the link and creates the file outside the jail.

Fix: replace `probe.exists()` with `probe.symlink_metadata().is_ok()` (lstat — succeeds on a dangling symlink). Then the dangling link becomes the "existing" probe, `canonicalize()` on it fails, and the existing `map_err(|_| WikiError::OutsideWiki)` refuses the path.

Constraints:
- Keep all existing tests green, including `refuses_a_symlink_pointing_out_of_the_wiki` and `allows_a_page_that_does_not_exist_yet`.
- Add a regression test in the same style as `refuses_a_symlink_pointing_out_of_the_wiki`: create a symlink inside the wiki whose target does **not** exist, and assert both `resolve` and `write_page` refuse it with `WikiError::OutsideWiki` — and that the outside target was not created.
- Unix-only test gating (`#[cfg(unix)]`) matching the existing symlink test.

</details>

src-tauri/src/wiki.rs
```diff
@@ -0,0 +1,937 @@
...
+
+    for entry in entries.flatten() {
+        let path = entry.path();
+        let name = entry.file_name();
+        let name = name.to_string_lossy();
+        // `.git` is the wiki's history, not its content.
+        if name.starts_with('.') {
+            continue;
+        }
+        if path.is_dir() {
```
positonic, line 284: **[Medium] `collect_pages` follows directory symlinks — a link cycle crashes the app, an outward link leaks listings**

`path.is_dir()` follows symlinks. `ln -s . loop` inside the wiki (easy to create by accident) makes `collect_pages` recurse forever until the stack blows — aborting the whole Tauri process on every `wiki_list_pages` *and* every `wiki_search`, since search lists first. A symlink to an outside directory is quieter but contradicts the jail: its `.md` filenames and sizes get listed into the prompt (reads are refused by `resolve`, but the names leak).

Suggested fix — skip symlinks entirely; `DirEntry::file_type()` is lstat-based and doesn't follow:

```suggestion
        if name.starts_with('.') {
            continue;
        }
        // Symlinks are skipped outright: following a directory link would leak
        // listings from outside the jail, and a link cycle (`ln -s . loop`)
        // would recurse until the stack blows.
        if entry.file_type().map(|t| t.is_symlink()).unwrap_or(true) {
            continue;
        }
        if path.is_dir() {
```

<details>

In `src-tauri/src/wiki.rs`, `collect_pages` (around lines 255–283) recurses with `path.is_dir()`, which follows symlinks. Two bugs: (1) a symlink cycle inside the wiki (e.g. `ln -s . loop`) causes unbounded recursion and a stack-overflow abort of the whole app on `wiki_list_pages`/`wiki_search`; (2) a symlink to an outside directory gets its contents listed (filenames + sizes) into the model prompt even though `resolve` would refuse reading them.

Fix: before the `path.is_dir()` check, skip any entry that is a symlink, using `entry.file_type()` (lstat-based, does not follow links); treat a `file_type()` error as skip.

Constraints:
- Keep the existing dot-entry skip and the `.md`-only filter as they are.
- Keep all existing tests green (`lists_markdown_pages_including_nested_ones`, ordering, search tests).
- Add regression tests (Unix-gated like the existing symlink test): a `loop` symlink pointing at the wiki root must not hang or crash `list_pages`, and a symlink to an outside directory containing a `.md` file must not appear in `list_pages` output.

</details>

src/lib/chat/streamLocalWikiChat.ts
```diff
@@ -0,0 +1,237 @@
...
+  // and must not be silently auto-retried.
+  let rawLength = 0;
+
+  const snapshot = (): ChatStreamUpdate => ({
+    displayText,
+    toolCalls: Array.from(toolCallsById.values()),
+    rawLength,
+  });
+
+  const response = await agent.stream(messages, { clientTools });
```
positonic, line 160: **[Medium] No idle timeout and no abort path — a silently stalled stream hangs the chat forever**

`streamChatResponse` carries a 60s idle watchdog precisely because "the server stream stalls silently"; this transport has neither that nor `AbortSignal` support. A Mastra stall mid-turn leaves `isStreaming` stuck with no recovery short of a reload — and since the minted token expires in 10 minutes, a long multi-round turn can also die mid-stream with nothing to bound the wait. The existing `classifyStreamError` already maps `'stream-idle-timeout'` to a retryable failure, so the recovery path is free once the watchdog throws.

Suggested fix: add an idle watchdog that is reset on every `onChunk` and rejects the turn with `new DOMException('stream-idle-timeout', 'AbortError')` when it fires (race it against `processDataStream`), and thread an optional `signal` through `StreamLocalWikiOptions` for parity with `StreamChatOptions`. If the pinned `@mastra/client-js` build accepts an abort signal, pass it through; if not, the race alone restores the UI.

<details>

In `src/lib/chat/streamLocalWikiChat.ts`, `streamLocalWikiChat` awaits `agent.stream(...)` and `processDataStream(...)` with no idle timeout and no `AbortSignal` support. If the Mastra stream stalls silently, the promise never settles and ManyChat's `isStreaming` is stuck forever. The sibling transport `src/lib/chat/streamChatResponse.ts` implements the pattern to copy: a 60s idle timer reset on every chunk that aborts with `new DOMException('stream-idle-timeout', 'AbortError')`, which `classifyStreamError` in `src/lib/chat/streamProtocol.ts` already classifies as a retryable `idle-timeout`.

Fix: add `idleTimeoutMs` (default 60_000) and optional `signal` to `StreamLocalWikiOptions`; reset the timer in `onChunk`; race `processDataStream` against a promise that rejects with the AbortError when the timer fires or the external signal aborts; always clear the timer in a `finally`.

Constraints:
- Don't change the function's public signature beyond adding optional fields to `StreamLocalWikiOptions`.
- The commit callback (`onTurnWrote`) must NOT run when the turn is aborted by the watchdog — the existing "commits after the stream ends" ordering must hold.
- Keep all existing tests in `src/lib/chat/__tests__/streamLocalWikiChat.test.ts` green, and add one using fake timers: a stream that emits nothing for longer than `idleTimeoutMs` rejects with a message containing `stream-idle-timeout`.

</details>

src/app/_components/ManyChat.tsx
```diff
@@ -1256,23 +1270,55 @@ export default function ManyChat({ initialMessages, githubSettings, buttons, pro
...
+              coreMessages,
+              buildWikiClientTools(wikiBridge),
+              {
+                onUpdate: renderUpdate,
+                // The user's own question reads better in `git log` than
+                // anything we could synthesise from the diff.
+                turnSummary: messageToSend,
+                onTurnWrote: async (summary) => {
+                  await wikiBridge.commitTurn(summary);
+                },
```
positonic, line 1317: **[Medium] A failed commit converts a fully-successful answer into "ended early" — and on git-less machines, *every* writing turn fails**

`commitTurn` runs after the stream has finished; if it rejects, the throw propagates out of `streamLocalWikiChat`, the final result is discarded, and the catch block marks the turn `incomplete` with a Retry footer — even though the answer streamed fully and the writes are on disk. Pressing Retry then re-runs the whole turn (re-spends tokens, re-executes writes) just to reattempt a commit. This isn't only an edge case: `wiki_init` deliberately tolerates a machine without git (`WikiInfo.git: false`), but on such a machine `commit_turn`'s `git add -A` fails, so every writing turn presents as broken. (Note also that Tauri `invoke` rejects with a plain **string**, which `classifyStreamError` maps to `unknown`.)

Suggested fix — fail the commit softly; uncommitted changes are swept into the next successful turn's `git add -A` anyway:

```suggestion
                onTurnWrote: async (summary) => {
                  try {
                    await wikiBridge.commitTurn(summary);
                  } catch (commitError) {
                    // The answer streamed and the writes are on disk; a failed
                    // commit (git missing, lock contention) must not convert
                    // the turn into "ended early". The next successful writing
                    // turn's `git add -A` sweeps these changes into its commit.
                    console.error('[ManyChat] wiki commit failed', commitError);
                  }
                },
```

<details>

In `src/app/_components/ManyChat.tsx` (around line 1315), the local-wiki branch passes `onTurnWrote: async (summary) => { await wikiBridge.commitTurn(summary); }` to `streamLocalWikiChat`. `streamLocalWikiChat` (in `src/lib/chat/streamLocalWikiChat.ts`) awaits this callback *after* the stream completes, so a rejected commit throws away an already-successful turn: ManyChat's catch block marks the bubble `incomplete` ("ended early") and its Retry re-runs the entire turn including the file writes. On a machine without git (`wiki_init` reports `git: false` and the wiki is documented to still work), the `git add -A` inside `wiki_commit_turn` fails, so every writing turn hits this.

Fix: wrap the `commitTurn` call in try/catch inside the `onTurnWrote` callback; on failure, log via `console.error` and do not rethrow. Optionally surface a subtle non-blocking notice, but do not mark the turn failed.

Constraints:
- Don't change `streamLocalWikiChat`'s contract (commit still only after the stream ends, only when a write succeeded).
- Keep the existing tests in `src/lib/chat/__tests__/streamLocalWikiChat.test.ts` green.
- Match the surrounding code style (comments explaining the *why*, existing log prefix conventions).

</details>

positonic (commented):
Changes needed

PR #527, opened 2026-08-06
fix(ticketSync): stop the body-repair pass deleting human Notion content

positonic:
> [!NOTE]
> ## Migration ordering
>
> Merging **does** apply the migration: `.github/workflows/prisma-migrate.yml`
> runs `prisma migrate deploy` against `DATABASE_URL_PRODUCTION` on any push to
> `main` touching `prisma/migrations/**`, which this PR does.
>
> What is *not* guaranteed is the ordering against the Vercel deploy — both start
> on merge and race. If the deploy lands first, `runOutboundCreate` creates the
> Notion page and then throws on the missing column, leaving an orphaned page.
> The `NonRetryablePushError` added here stops that from being retried, so the
> exposure is one orphan per affected ticket rather than `MAX_ATTEMPTS`, and only
> for tickets needing a *new* page on a push-enabled connection during the window.
>
> To close the race entirely, run against the production `DATABASE_URL` before merging:
>
> - [ ] `npx prisma migrate status` — confirm `20260806090000_ticket_sync_remote_created_at` is pending
> - [ ] `npx prisma migrate deploy` (**never** `migrate dev` against production)
> - [ ] `npx prisma migrate status` — confirm "Database schema is up to date"
>
> The workflow then no-ops on merge. Optional, but it costs nothing.

## The incident

The body-repair pass added in PR 521 / PR 522 deleted human-authored content
from Notion pages in the CLEAR ticket database, replacing it with a bare
"Synced from Exponential" callout. Affected pages lost content that had never
been copied into Exponential, so it existed nowhere.

## Root cause

`bodyRepair` decided "we created this Notion page" by scanning the run ledger
for `items[].action === "created"`. The **inbound** engine writes that same
token with the opposite meaning — "created a *ticket* from this page" — and
the ledger query filtered only on `configId`, never on `direction`. Every page
ever imported from Notion was therefore classified as machine-authored and had
its top-level blocks deleted.

Where the ticket had no body, `buildBodyBlocks` emits only the callout, so the
page was left with the callout alone. Inbound copies page content only on a
row's first import and never on update, so anything written in Notion after
import had never reached Exponential either.

Inconsistent blast radius came from three unrelated filters — ledger presence,
whether the ticket happened to have a body, and how far the batch cursor got —
none of which was "did we author this page".

`revert.ts` already gated correctly on `direction === "pull"`; `bodyRepair` was
the only consumer missing that check.

## The fix

Provenance is now a recorded fact rather than an inference, with three further
guards in front of the delete. Each is independently sufficient:

- **`TicketSync.remoteCreatedAt`** — stamped by the outbound create path at the
  moment it creates a page. The run ledger is no longer consulted at all; a
  test asserts `ticketSyncRun.findMany` is never called.
- **Non-empty body** — a ticket with no body is skipped. There is nothing to
  restore, so deleting would be pure loss.
- **Shape check** — the page's current blocks must still match what the
  creation path writes (sync callout + flat paragraphs, nothing nested). Any
  heading, to-do, attachment or child block means a human has been there.
  This guard alone would have prevented the incident.
- **Dry run by default** — writing requires an explicit opt-out, and the tRPC
  input defaults to `dryRun: true`.

Skipped pages are reported with a reason instead of being silently passed over.
Also removes `listBlockChildrenIds`: an ids-only listing made it too easy to
delete blocks without ever inspecting their types.

## Migration

Adds a nullable `TicketSync.remoteCreatedAt` and backfills it from
**push-direction, non-dry-run** runs only. Additive and backward compatible
with the currently deployed code, so it should be applied **before** this
merges — `npm run build` does not run migrations, and the new push path writes
the column.

Validated against a throwaway Postgres with fixture rows: the push-created page
is stamped (earliest run wins), while inbound-`created`, dry-run, and
same-page-different-config rows correctly stay `NULL`. Null, non-array, and
missing-`externalId` ledger entries do not error.

## Testing

Full unit suite green (2284 tests) and `tsc --noEmit` clean. 14 tests on
`bodyRepair`, covering each guard, plus a `push` test pinning the provenance
stamp at its source.


prisma/migrations/20260806090000_ticket_sync_remote_created_at/migration.sql
```diff
@@ -0,0 +1,31 @@
+-- Explicit provenance for Notion pages the OUTBOUND push created.
+--
+-- Until now the only record of "we authored this page" was the run ledger's
+-- `items[].action === "created"` — which the INBOUND engine also writes, with
+-- the opposite meaning ("created a ticket from this page"). Reading that as
+-- page provenance let the body-repair pass wipe human-authored Notion pages.
+ALTER TABLE "TicketSync" ADD COLUMN "remoteCreatedAt" TIMESTAMP(3);
```
positonic, line 7: **[High] The migration has no enforcement, and merging ahead of it duplicates Notion pages**

`vercel.json`'s `buildCommand` is `npm run build` → `next build` only; nothing runs `prisma migrate deploy`. If this merges before the migration is applied, the failure is not "the repair pass is unavailable" — it's active damage on the push path:

1. `runOutboundCreate` calls `adapter.createPage(...)` — the Notion page is really created.
2. The follow-up `db.ticketSync.update` now writes `remoteCreatedAt` → `column "remoteCreatedAt" does not exist` → throws.
3. The sync row keeps its `pending:<ticketId>` sentinel, so `pushRunner` retries the job — up to `MAX_ATTEMPTS = 5`, each attempt creating **another** page.
4. Every orphan is un-marked by a sync record, so the `*/10` inbound cron re-imports them as new tickets.

So the window between merge and migration produces up to 5 duplicate Notion pages per created ticket plus their inbound re-imports. The PR description asks for the migration to be applied first, but a sentence in a description is not a mechanism.

Suggested fix: make the ordering structural rather than procedural — either gate the merge on the migration being applied to the production DB (and say so in the PR checklist), or land the column write behind a guard so a missing column degrades instead of throwing. The cheapest structural option is to split this into two merges: migration-only first, then the code.

<details>

This repo deploys via Vercel with `buildCommand: npm run build` (`next build`), and no step runs `prisma migrate deploy`. The PR adds `prisma/migrations/20260806090000_ticket_sync_remote_created_at/migration.sql` (a new nullable `TicketSync.remoteCreatedAt` column) and, in `src/server/services/ticketSync/push.ts` around line 622, starts writing that column from `runOutboundCreate`.

Problem: if the code deploys before the migration is applied, `runOutboundCreate` will have already called `adapter.createPage` (real Notion page created) and then throw on the `db.ticketSync.update`. `pushRunner.ts` retries the job up to `MAX_ATTEMPTS = 5`, and each retry re-enters the create branch because the sync row still holds its `pending:<ticketId>` sentinel — producing duplicate Notion pages that the inbound cron then re-imports as duplicate tickets.

Do two things:
1. Document the required ordering as an explicit pre-merge step in the PR body / a deploy checklist, naming the exact command and target database.
2. Make the code tolerate the pre-migration state OR make the ordering structural (e.g. split into a migration-only PR that merges and deploys first, then the code PR).

Constraints:
- Do not change `MAX_ATTEMPTS` or the retry/backoff semantics as a workaround.
- Do not make `remoteCreatedAt` non-nullable.
- Keep all existing tests passing.

</details>

src/server/services/ticketSync/push.ts
```diff
@@ -609,13 +609,17 @@ async function runOutboundCreate(
...
+  // machine-authored — the run ledger's "created" action is ambiguous across
+  // directions. Anything that rewrites page CONTENT must gate on this column.
   await db.ticketSync.update({
     where: { id: sync.id },
     data: {
       externalId,
       externalUrl: url,
       snapshot: local as unknown as Prisma.InputJsonValue,
       lastSyncedAt: new Date(),
+      remoteCreatedAt: new Date(),
```
positonic: **[Medium] A failed post-`createPage` DB write is retried, and the retry creates a second page**

This update is the *only* record that the page we just created belongs to this sync — and it runs after the irreversible remote write. If it throws for any reason (this new column pre-migration, a connection drop, a serverless freeze), the sync row keeps its `pending:<ticketId>` sentinel, `pushRunner` marks the job `PENDING` with backoff, and the next drain re-enters `runOutboundCreate` and creates a **second** Notion page. Up to `MAX_ATTEMPTS = 5` of them, all orphaned and all re-importable by the inbound poll.

Retrying is the wrong response to a failure that happens *after* the remote side already succeeded. Pre-existing, but this PR makes the write carry provenance, so its failure is now strictly more expensive.

Suggested fix: treat a post-create bookkeeping failure as terminal rather than retryable, so the operator gets one orphan to reconcile instead of five.

```ts
  const { externalId, url } = await adapter.createPage({ ... });

  try {
    await db.ticketSync.update({ where: { id: sync.id }, data: { ... } });
  } catch (error) {
    // The page exists remotely; retrying would create a duplicate. Fail hard.
    throw new NonRetryablePushError(
      `Notion page ${externalId} was created but could not be linked: ${
        error instanceof Error ? error.message : "unknown error"
      }`,
    );
  }
```

<details>

In `src/server/services/ticketSync/push.ts`, `runOutboundCreate` calls `adapter.createPage(...)` and then does a `db.ticketSync.update` (around line 616-624) to replace the `pending:<ticketId>` sentinel with the real `externalId` and stamp `remoteCreatedAt`.

Bug: if that update throws, the Notion page already exists but the sync row still holds the sentinel. `src/server/services/ticketSync/pushRunner.ts` catches the throw, leaves the job `PENDING` with backoff, and the next drain re-enters the create branch — creating a duplicate Notion page, up to `MAX_ATTEMPTS = 5` times. Each orphan is later re-imported by the inbound sync as a new ticket.

Make a post-`createPage` bookkeeping failure terminal instead of retryable:
- Introduce a distinguishable error (e.g. a `NonRetryablePushError` class) thrown when the link-update fails after the remote page exists.
- In `pushRunner.ts`, when that error type is caught, set the job to `FAILED` immediately (do not schedule a retry) and record a run item whose `reason` names the orphaned Notion page id so an operator can reconcile it.

Constraints:
- Do not change `MAX_ATTEMPTS`, `backoffMs`, or the behavior for ordinary transient failures (rate limits, 5xx) — those must still retry.
- Do not change `runOutboundCreate`'s return type or the `OutboundPushItem` shape.
- Add a unit test in `src/server/services/ticketSync/__tests__/push.test.ts` (or `pushRunner`'s suite) proving that a throwing `ticketSync.update` after a successful `createPage` does not lead to a second `createPage` call.
- Keep the existing 2284-test suite green.

</details>

src/server/services/ticketSync/bodyRepair.ts
```diff
@@ -63,22 +78,44 @@ type NotionFactory = (
...
+      return typeof text?.content === "string" ? text.content : "";
+    })
+    .join("");
+}
+
+/**
+ * Does this page's content still look like something the creation path wrote?
+ *
+ * The creation path emits exactly: the sync callout, then flat paragraph
+ * blocks — nothing nested, no other block types. That signature is cheap to
```
positonic: **[Medium] The shape guard matches the *legacy* creation path, so today's pages are reported as human-edited**

"The creation path emits exactly: the sync callout, then flat paragraph blocks" is true of the pre-#521 builder (`git show 2baf9636:…/outboundCreate.ts` — split on `\n{2,}`, one paragraph each), but not of the current one: `buildBodyBlocks` → `markdownToNotionBlocks` emits `heading_1..3`, `code`, `bulleted_list_item`, `numbered_list_item`, `quote` and `divider`.

Two consequences:

1. **The reported reason is wrong.** Any page created after PR 521 whose body has a heading or a list gets skipped with `"page content has been edited in Notion — left untouched to avoid deleting human work"`. Nobody edited it; we wrote it that way. An operator triaging a sweep is being told something false about a customer's workspace.
2. **The pass is one-shot.** A page it repairs no longer matches the guard, so it can never be repaired again. That's fine today, but the docstring gives a future maintainer exactly the wrong reason for the mismatch — and "fix" the guard by admitting headings/lists and the incident window reopens (a human-edited page with headings would then pass).

The guard's tightness is correct. Only the description and the skip reason need to be honest about *why*.

Suggested fix:

```suggestion
 * The LEGACY creation path — the one whose pages this sweep exists to fix —
 * emitted exactly: the sync callout, then flat paragraph blocks. Today's path
 * renders real Markdown (headings, lists, code), so a page written by the
 * current renderer deliberately does NOT match: this pass is one-shot by
 * design, and the guard must stay this tight. That signature is cheap to
```

<details>

In `src/server/services/ticketSync/bodyRepair.ts`, the docstring on `isSyncAuthoredContent` (around lines 94-103) and the module doc's guard 3 (around line 26) both claim the check matches "what the creation path writes". That is only true of the legacy creation path, which emitted the sync callout plus flat paragraphs. The current `buildBodyBlocks` → `markdownToNotionBlocks` emits `heading_1..3`, `code`, `bulleted_list_item`, `numbered_list_item`, `quote` and `divider`.

Fix the documentation and the operator-facing message, not the guard:

1. Reword both docstrings to say the check matches the LEGACY creation path (the population this sweep exists to repair), and state explicitly that a page written by the current renderer will not match — the pass is one-shot by design and the guard must stay this tight, because loosening it to accept headings/lists would re-admit human-edited pages.
2. Around line 240, split the skip reason. When the blocks do not lead with the sync callout at all, keep "page content has been edited in Notion". When they lead with the sync callout but contain non-paragraph blocks below it, use a distinct reason such as "already rendered as Notion blocks — nothing to repair", since that is the current renderer's own output, not a human edit.
3. Add a unit test asserting that a page shaped like today's renderer output (callout + `heading_2` + `bulleted_list_item`) is skipped with the "nothing to repair" reason, not the human-edit one.

Constraints:
- Do NOT loosen `isSyncAuthoredContent` to accept headings, lists, code, quotes or dividers. Its tightness is the guard that would have prevented the incident.
- Keep `isSyncAuthoredContent`'s signature and return type unchanged.
- Keep the existing 14 bodyRepair tests passing.

</details>

src/server/services/ticketSync/bodyRepair.ts
```diff
@@ -166,32 +204,67 @@ export async function rerenderCreatedPageBodies(
...
+        ...itemBase,
+        outcome: "skipped",
+        reason:
+          "ticket has no body — refusing to replace the page content with an empty one",
+      });
+      continue;
+    }
+
     try {
+      const existing = await notion.listBlockChildren(sync.externalId);
```
positonic, line 275: **[Medium] `last_edited_by` vs the bot id is the authoritative human-edit signal, and it's already plumbed into this function**

Guard 3 infers "has a human been here?" from block shape. Notion answers the question directly, and this repo already relies on that answer: `notionAdapter.ts:198` computes `lastEditedByBot` as `page.last_edited_by?.id === this.botId`, and ADR-0046 names it as the echo-suppression mechanism. `NotionFactory` in this very file already declares `botId: string | null` in its success branch (line 77) — and nothing reads it.

A page whose `last_edited_by` is not our bot has been touched by a person, whatever its blocks look like. That catches the case the shape check cannot: a human adding *plain paragraphs* to a sync-created page, which currently passes every guard and gets deleted.

Suggested fix — one extra API call per candidate, strictly narrowing:

```ts
      const page = await notion.getPage(sync.externalId);
      const lastEditor = (page as { last_edited_by?: { id?: string } })
        .last_edited_by?.id;
      if (botId && lastEditor && lastEditor !== botId) {
        skipped++;
        items.push({
          ...itemBase,
          outcome: "skipped",
          reason: "page was last edited by a person, not the sync bot",
        });
        continue;
      }

      const existing = await notion.listBlockChildren(sync.externalId);
```

<details>

In `src/server/services/ticketSync/bodyRepair.ts`, `rerenderCreatedPageBodies` guards a destructive block-delete with a block-shape heuristic (`isSyncAuthoredContent`). Add an additional, stronger guard using Notion's own `last_edited_by`.

Facts you need:
- The `NotionFactory` type in this file already returns `botId: string | null` on success (around line 77); the value is currently discarded — capture it from `resolved`.
- `src/server/services/ticketSync/notionAdapter.ts:198` already derives `lastEditedByBot` as `!!this.botId && page.last_edited_by?.id === this.botId`. Mirror that comparison exactly.
- `NotionService.getPage(pageId)` exists and returns the page object including `last_edited_by`.

Changes:
1. Add `getPage(pageId: string): Promise<Record<string, unknown>>` to the local `NotionBlockOps` interface.
2. Capture `botId` from the factory result alongside `notion`.
3. Inside the per-sync `try`, before `listBlockChildren`, fetch the page and skip (outcome `"skipped"`, reason naming the human editor) when `botId` is non-null and `last_edited_by.id` differs from it.
4. Treat an unknown/absent `botId` or missing `last_edited_by` as "cannot tell" and fall through to the existing shape check — never as permission to write.

Constraints:
- This guard must only ever ADD skips. Do not weaken or remove `isSyncAuthoredContent`, the non-empty-body guard, or the dry-run default.
- Update the module docstring's numbered guard list to include it.
- Add unit tests: (a) a page whose `last_edited_by.id` differs from `botId` is skipped even when its blocks are callout + paragraphs; (b) a page whose `last_edited_by.id` equals `botId` still repairs; (c) `botId: null` falls through to the existing behavior (the current test fake passes `botId: null`, so all 14 existing tests must stay green unchanged).

</details>

src/server/services/ticketSync/bodyRepair.ts
```diff
@@ -166,32 +204,67 @@ export async function rerenderCreatedPageBodies(
...
+
       const blocks = buildBodyBlocks(sync.ticket.body, backlinkUrl);
-      const existing = await notion.listBlockChildrenIds(sync.externalId);
-      for (const blockId of existing) {
-        await notion.deleteBlock(blockId);
+      for (const block of existing) {
+        const blockId = typeof block.id === "string" ? block.id : null;
+        if (blockId) await notion.deleteBlock(blockId);
       }
       await notion.appendBlockChildren(sync.externalId, blocks);
```
positonic, line 309: **[Medium] Delete-then-append is not crash-safe; a failed append leaves the page empty**

Blocks are trashed first, then the replacement is appended. If the append fails — 429, a validation rejection, or the serverless function timing out mid-loop — the page is left **empty**, and the operator sees only `outcome: "failed"` with the raw Notion error. Nothing in the report says "this page currently has no content".

`markdownToNotionBlocks` caps at `MAX_BLOCKS = 90`, plus the callout is 91, and `appendBlockChildren` chunks at 90 — so a long body is two append calls and a partial write is reachable, not theoretical.

Since we hold the exact ids to remove, the safe order is free: append first, then delete. Worst case is a page with duplicated content (visible, recoverable, obvious) instead of an empty one.

Suggested fix:

```suggestion
      const blocks = buildBodyBlocks(sync.ticket.body, backlinkUrl);
      // Append BEFORE deleting: a failure here leaves the page with duplicate
      // content (visible and recoverable), never with no content at all.
      await notion.appendBlockChildren(sync.externalId, blocks);
      for (const block of existing) {
        const blockId = typeof block.id === "string" ? block.id : null;
        if (blockId) await notion.deleteBlock(blockId);
      }
```

<details>

In `src/server/services/ticketSync/bodyRepair.ts`, `rerenderCreatedPageBodies` currently deletes every existing top-level block and then appends the freshly rendered ones (around lines 256-261). If the append fails after the deletes have run, the Notion page is left with no content and the item is reported only as `outcome: "failed"`.

Swap the order: append the new blocks first, then delete the previously-listed block ids. The ids were captured before the append, so the new blocks are never in the delete set. A failure then leaves duplicated content rather than an empty page.

Also make the failure legible: when a delete fails after the append succeeded, the item's `reason` should say the page now holds both the old and new content and needs manual cleanup.

Constraints:
- Do not change the guards ordering — the provenance, non-empty-body, shape and dry-run checks must all still run before any write.
- Do not change `BodyRepairResult`'s shape.
- Add a unit test where `appendBlockChildren` rejects, asserting the existing blocks were NOT deleted and the item is reported as `failed`.
- Keep the existing 14 bodyRepair tests green (note the current "continues past a failing page" test fails on `listBlockChildren`, which is unaffected by this reorder).

</details>

src/server/services/ticketSync/bodyRepair.ts
```diff
@@ -166,32 +204,67 @@ export async function rerenderCreatedPageBodies(
...
+        });
+        continue;
+      }
+
+      if (dryRun) {
+        skipped++;
+        items.push({
+          ...itemBase,
+          outcome: "skipped",
+          reason: `dry run — would replace ${existing.length} block(s) with the re-rendered body`,
```
positonic: **[Low] The dry run reports as `skipped`, so the preview can't be read as a plan**

A dry run emits `outcome: "skipped"` for pages it *would* repair — the same outcome as pages a guard *refused*. So `{ repaired: 0, skipped: 12 }` could mean twelve pages queued for rewrite, twelve pages protected, or any mix; only free-text `reason` prefixes distinguish them, and only by string matching.

That undercuts the point of making dry run the default: an operator can't compare the preview against the real run's counts to confirm the sweep did what it said it would.

Suggested fix — a distinct outcome, so counts are comparable across the two runs:

```ts
      if (dryRun) {
        wouldRepair++;
        items.push({
          ...itemBase,
          outcome: "would-repair",
          reason: `would replace ${existing.length} block(s) with the re-rendered body`,
        });
        continue;
      }
```

<details>

In `src/server/services/ticketSync/bodyRepair.ts`, the dry-run branch (around lines 246-253) pushes `outcome: "skipped"` for pages it would have repaired, and increments the same `skipped` counter used for pages a guard refused. A caller reading `BodyRepairResult` cannot tell "would rewrite" from "protected" without string-matching the `reason` field.

Separate them:
1. Add `"would-repair"` to `BodyRepairItem["outcome"]`.
2. Add a `wouldRepair: number` field to `BodyRepairResult`, incremented only in the dry-run branch (and always present, 0 on a live run).
3. Keep `skipped` for genuine guard refusals only.
4. Drop the now-redundant "dry run — " prefix from the reason string; the outcome carries it.

Constraints:
- Do not change the dry-run DEFAULT (`params.dryRun !== false`) or the tRPC input default of `true` in `src/plugins/product/server/routers/ticketSync.ts`.
- Update the tests in `src/server/services/ticketSync/__tests__/bodyRepair.test.ts` — note the "is a no-op when no page was created by the push" test asserts the full result object with `toEqual`, so it needs the new field.
- Keep all other bodyRepair tests green.

</details>

positonic (commented):
The fix itself is sound — stamping provenance at the moment of creation instead of inferring it from an ambiguous ledger token is the right call, and the shape check is a good last line of defence that would have caught the incident on its own.

Nothing in the diff is broken. My concerns are about **how this lands** and one adjacent defect it makes more expensive:

- **Blocking-ish:** the migration must be applied before this merges, and nothing enforces that. The failure mode isn't "feature unavailable" — `createPage` succeeds, the column write throws, the job retries 5×, and you get up to 5 duplicate Notion pages per ticket plus their inbound re-imports.
- **Worth fixing here:** a post-`createPage` DB failure is currently retried, which is what turns the above into duplicates rather than one clean error.
- **Worth a follow-up:** `last_edited_by` vs the bot id is an authoritative human-edit signal that this file's own factory type already carries and never reads; delete-then-append can leave a page empty; and the shape guard's docstring describes the legacy creation path as if it were the current one.

Details and paste-ready fix prompts are in the inline comments.

---
## Adversarial production-risk review

The core fix is right: replacing the ledger inference with a stamped `TicketSync.remoteCreatedAt` is the correct call, the shape check would genuinely have prevented the incident on its own, and the guard layering is honest about what each layer buys. The findings are about the edges — the deploy window, the crash-safety of the delete, and a stronger provenance signal that is already plumbed in but unused.

**6 inline findings: 1 High, 4 Medium, 1 Low.** Worst issue: nothing enforces the migration landing before this merges, and the failure mode is duplicate + orphaned Notion pages in the customer's workspace, not a disabled feature.

Not verified by me: the "2284 tests green, `tsc --noEmit` clean" claim — the review worktree had no `node_modules`. Every finding below and inline was confirmed by reading the post-change source directly.

---

### [Low] `CONTEXT.md` still names the run ledger as the ground truth this PR just disqualified

`CONTEXT.md`'s **Sync run** entry (line 206) reads: *"The run record is the only ground truth for 'this ticket was created by sync'"*. That sentence is about ticket provenance rather than page provenance, but it is the exact conflation that produced this incident — and it's the sentence a future maintainer will find when they ask "how do I tell whether we made this?"

Meanwhile `remoteCreatedAt` — the new answer for pages — appears nowhere outside the schema comment and `bodyRepair.ts`. `CLAUDE.md` asks for glossary and ADR updates *inline with the change*, not batched.

Two small additions would close it:

- **`CONTEXT.md`, Sync run entry:** narrow the ledger claim to *tickets*, and point page provenance at `TicketSync.remoteCreatedAt`.
- **`CONTEXT.md`, Outbound push entry:** it already ends with "There is no revert for pushed changes" — the natural home for "page CONTENT is machine-writable only where `remoteCreatedAt` is set; a null link is human-authored and never rewritten."

Whether this also warrants an ADR is a judgement call. ADR-0046's Consequences section already predicted this exact class of failure ("a wrong write reaches the customer's live Notion workspace with no revert story"), and it now has: a hard-to-reverse rule, a real rejected alternative (ledger inference), and an incident behind it. A short amendment to ADR-0046 rather than a new ADR would keep the trail in one place.

**Context for the reviewer:** this is the third consumer of "who created this" in the sync subsystem — `revert.ts` gates on `direction === "pull"`, `bodyRepair` now gates on the column, and the glossary says the ledger. Two of the three are right; the doc is the one that isn't.

PR #544, opened 2026-08-11
feat(auth): sign-in codes replace magic links

positonic:
## What

Email sign-in delivers an 8-character Crockford base32 **Sign-in code** that the user types, and no link at all.

Corporate mail scanners (Mimecast URL Protect and equivalents) rewrite and follow URLs in email, and a NextAuth email token is single-use, so the scanner spends the token and the human gets "invalid or expired link". Observed against a real user at the Norwegian Refugee Council. For that cohort the failure is total: they cannot sign in, and because they cannot sign in they cannot create an account either.

The code **is** the NextAuth verification token — `generateVerificationToken` replaces Auth.js's `randomString(32)`, so redemption runs through the ordinary `/api/auth/callback/postmark` and user creation, Personal-workspace bootstrap, pending-invite acceptance and `emailVerified` all stay on the standard path. **No migration** — this reuses `VerificationToken`.

Security rests on entropy, not throttling: verification lands on Auth.js's own callback, which has no rate limiting and is excluded from the `middleware.ts` matcher. Hence 8 base32 characters (~10^12) rather than a 6-digit PIN.

Decisions recorded in [ADR-0056](docs/adr/0056-sign-in-codes-replace-magic-links.md) (this change) and [ADR-0055](docs/adr/0055-password-credential-in-its-own-table.md) (the V2 decision that depends on it).

## Commits, one per ticket action

1. **Returning user signs in with an emailed code** — `src/lib/signInCode.ts` + 11 unit tests, `generateVerificationToken`/`maxAge` on the Postmark provider, `sendSignInCodeEmail`, the verify-page code form, the `/signin` handoff
2. **New-user path carries a code, not a link** — the welcome email renders the code instead of a CTA button
3. **Invite and applicant-account surfaces hand off to the code form** — `/invite/[token]` and the public form's applicant CTA
4. **Failure paths explain themselves** — Auth.js `?error=` mapping, send-failure handling on both surfaces
5. **E2E coverage at the HTTP seam** — three Playwright tests

## Acceptance criteria

- [x] The sign-in email contains an 8-character Crockford base32 code and **no** sign-in URL
- [x] A code expires 10 minutes after issue and is stored hashed
- [x] Codes are accepted irrespective of case, the display hyphen and surrounding whitespace; typed `I`/`L`/`O` are read as `1`/`1`/`0`
- [x] Redeeming a valid code creates the User, their Personal workspace and any pending invitation acceptances
- [x] An incorrect, expired or used code returns the user to `/signin` with a message naming the cause
- [x] A failed send never tells the user to check their inbox
- [ ] No email address appears in a URL query parameter — **partially met, see below**
- [x] Codes are drawn from a space of at least 10^12 values

### The one criterion not fully met

The `/signin` → verify-page handoff goes through `sessionStorage`, and no email address appears in any link we mail. But redemption is a navigation to `/api/auth/callback/postmark?token=...&email=...`, because Auth.js reads the identifier from `query.email` and has **no POST path** for email providers (`@auth/core/lib/actions/callback/index.js` reads `query.token`/`query.email`; `lib/utils/web.js` only populates `body` for POST, which the email branch never reads).

So the address lands in browser history and the server access log — the same exposure the magic link already had, no better and no worse. Closing it fully means a bespoke route that POSTs, builds the callback URL server-side and forwards the Auth.js response — a hand-rolled proxy in front of the one path ADR-0056 deliberately keeps stock. Not done here; worth a follow-up if the requirement is hard.

## Beyond the ticket, and why

- **The invite page had no send-failure branch at all** — it routed to "check your email" whether or not the email left. Fixed; shares the message constant with `/signin`.
- **`EmailSignin` is not in Auth.js's `clientErrors` set**, so a failed send reaches the browser as `?error=Configuration`. Mapping `EmailSignin` in the error table was dead code. Both surfaces now catch the failure at the call site, where the cause is still known.
- **`/f/[slug]`'s applicant-account CTA** was a third caller of `signIn("postmark")`, not listed in the actions. Left alone it would have handed applicants a code with nowhere to type it.

## Testing

- 11 unit tests on the pure functions (alphabet membership, absence of `I`/`L`/`O`/`U`, full 32-character coverage over 2000 draws, normalize/format round-trip)
- 3 Playwright tests over real HTTP: new-user redemption (session cookie + `emailVerified` + Personal workspace + invite acceptance + token consumed), fresh-tab email fallback + wrong-code messaging, and the `/signin` handoff + failed send
- No test triggers a real Postmark send: two seed the `VerificationToken` directly, the third intercepts the request in the browser
- Full unit suite green (2406 tests); full E2E suite green

---

Ships: outer.almond

[View in Exponential](https://www.exponential.im/w/syntrofi/tickets/cmsoqljru0007le04lcex06wo)

___

### **PR Type**
Enhancement, Tests, Documentation

___

### **Description**
- Replace magic links with typed 8-char sign-in codes

- Add `src/lib/signInCode.ts` helpers plus unit tests

- Rework `/signin`, invite, verify-request and public form UX

- Update code-bearing emails, retire older codes, add ADRs

___

### Diagram Walkthrough

```mermaid
flowchart TD
  signin["/signin, invite, public form"] -- "signIn(postmark, redirect:false)" --> provider["Postmark provider<br/>generateVerificationToken"]
  provider -- "8-char code, 10 min TTL" --> email["Sign-in code / welcome email"]
  provider -- "retire older tokens" --> db["VerificationToken"]
  email -- "user types code" --> form["SignInCodeForm on /auth/verify-request"]
  form -- "callback URL with token+email" --> callback["/api/auth/callback/postmark"]
  callback -- "events.createUser" --> session["Session + workspace bootstrap"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>8 files</summary><table>
<tr>
  <td><strong>signInCode.ts</strong><dd><code>New sign-in code generation, normalization and handoff helpers</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-f883092b25a9de0e1683d86685182d47461f9e478c5ca63ef4e53f98ae43a3df">+121/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config.ts</strong><dd><code>Postmark provider mints codes and retires older tokens</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-0e2d0a0ece50e4e7afb5174dbcef88e68ad9bfe47d53450bede288ed13084cd7">+75/-7</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>EmailService.ts</strong><dd><code>Emails render sign-in code instead of link</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-7efb63f19977ee348b1afbdb35fb8de2ac2bfa66d1cc9fcbb1c957a5959b7fcd">+83/-43</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>SignInCodeForm.tsx</strong><dd><code>New client form to redeem a sign-in code</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-4d10d47e41a6e93b02626b5c82713b03695f4f0091f9f1f61e279340b62952da">+104/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>page.tsx</strong><dd><code>Verify page hosts code form and new copy</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-93580fd4696db998511c14ffde93ebe22b052f0e702d6b2945442c968053c13c">+15/-10</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>SignInLandingPage.tsx</strong><dd><code>Handoff identifier, error messaging, send-failure handling</code></dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-b891ac2c13c8d32d160504490b6c6ab225d0569c6153058a69f1fa31df8de141">+74/-5</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>page.tsx</strong><dd><code>Invite page sends code and handles failures</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-73c22ad002c46aa25771d8e244d322edcbfb0f1c74a3fc518706a620f293bcba">+41/-4</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>PublicForm.tsx</strong><dd><code>Applicant account CTA hands off to code form</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-477d22ebaebb95d9f645f090403892e95bb1077f5f7ac500e1e0079843ef7c8e">+23/-4</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Formatting</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>auth-surface.css</strong><dd><code>Add `.auth-error` styling for sign-in failures</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-47d00a1d01cb843785813774e30f8f9827c7c08db5ee2b9f2d221d1a934ffda9">+11/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>page.tsx</strong><dd><code>Update metadata description wording</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-321f5935fac9f79a51848f042c6ea8b4964974c9dda277e5f5099dd8094f43db">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>0056-sign-in-codes-replace-magic-links.md</strong><dd><code>ADR documenting sign-in codes decision</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-db5b1e686b9a39ec7685164381dac0d95388056332b20aa6e9328bcd2a8c539b">+50/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>0055-password-credential-in-its-own-table.md</strong><dd><code>ADR for password credential table design</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-8de6c0df48d1118cef6c8e652ac61f8b6e8076be6769cf46529a933d2cc2f92a">+37/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>CONTEXT.md</strong><dd><code>Document Sign-in code and Password credential terms</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-86fee53c33308768d49a4ac0ff5720fd7f5411a8c8bd62ab9b80bbffa2977a14">+10/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>sign-in-code.spec.ts</strong><dd><code>End-to-end coverage for code redemption and handoff</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-1b7d190fe3c83e1911165c1dfceb1527cc3f050ed8c8de1b2d5f33bbab67cc6c">+268/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>signInCode.test.ts</strong><dd><code>Unit tests for alphabet, generation and normalization</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/positonic/exponential/pull/544/files#diff-5a2d4e7719f90895e08f135555c37745f0f3d5ba869ad49fd03a4717bfeea82d">+97/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr></tr></tbody></table>

</details>

___


src/server/auth/config.ts
```diff
@@ -196,29 +197,73 @@ export const authConfig = {
...
+          // silently diverges and this deletes the live row instead of the
+          // stale ones — sign-in breaks for everyone. The residual cost of the
+          // timestamp approach is far smaller: two near-simultaneous requests
+          // for one address, on instances whose clocks disagree by more than
+          // the gap between them, can retire the newer code instead of the
+          // older. The user asks for another code.
+          //
+          // Emails only: legacy API-key rows share this table under an
+          // `api-key:<userId>:<uid>` identifier, which never matches.
+          await db.verificationToken.deleteMany({
```
positonic, line 275: **[High] Retiring older codes turns unthrottled issuance into a targeted sign-in lockout**

`POST /api/auth/signin/postmark` is unauthenticated, has no throttle, and is excluded from the `middleware.ts` matcher (`'/((?!api|...).*)'`). Its CSRF token is free from `GET /api/auth/csrf`. Anyone who knows a victim's address can now loop that endpoint; each request retires the victim's live code, so the code sitting in their inbox is dead before they finish typing it. Sustained, a user with no OAuth account can never sign in. Before this commit N concurrent magic links all stayed valid, so this is a new availability regression — and each loop iteration also bills a Postmark send to a real inbox.

Suggested fix — throttle issuance per identifier before retiring anything (this repo already has the pattern in `WhatsAppVerificationService`: `MAX_CODES_PER_HOUR = 5`, `MAX_ATTEMPTS = 3`):

```ts
// Refuse to reissue while a recently-minted code is still live, so an
// attacker can't evict the code the user is holding.
const REISSUE_COOLDOWN_MS = 60_000;
const recent = await db.verificationToken.findFirst({
  where: {
    identifier,
    expires: { gt: new Date(expires.getTime() - REISSUE_COOLDOWN_MS) },
  },
});
if (recent) throw new Error("A sign-in code was already sent recently.");

await db.verificationToken.deleteMany({
  where: { identifier, expires: { lt: expires } },
});
```

<details>

In `src/server/auth/config.ts` around line 247, the Postmark provider's `sendVerificationRequest` deletes every earlier `VerificationToken` row for the identifier on each issue. Combined with `/api/auth/signin/postmark` having no rate limiting (it's excluded by the `middleware.ts` matcher and its CSRF token is obtainable from `GET /api/auth/csrf`), any unauthenticated caller can repeatedly request codes for a known address and permanently invalidate the one the real user is holding — a complete lockout for users without an OAuth account.

Add throttling on issuance keyed by identifier (and ideally by client IP too), modelled on `src/server/services/whatsapp/VerificationService.ts`, which already does this with `MAX_CODES_PER_HOUR = 5` and `MAX_ATTEMPTS = 3`. Requirements:

- Keep the existing retirement behaviour for legitimate reissues; only prevent an attacker from evicting a live code faster than the user can use it.
- Note that Auth.js runs `sendVerificationRequest` and `adapter.createVerificationToken` concurrently inside `Promise.all` (see `@auth/core/lib/actions/signin/send-token.js`), so throwing from `sendVerificationRequest` still leaves an orphan token row — make sure your approach either tolerates or cleans up that row.
- Consider whether the throttle belongs in `sendVerificationRequest` or in a wrapper around the Auth.js route handler / an added `middleware.ts` matcher entry for `/api/auth/signin/:path*` and `/api/auth/callback/postmark`; the route-level option also closes the unrated-limited guessing surface on the callback.
- Update `docs/adr/0056-sign-in-codes-replace-magic-links.md` and the **Sign-in code** entry in `CONTEXT.md`, both of which currently assert that safety rests on "entropy, not throttling" because the endpoint "cannot be throttled".
- Add a unit test (mocked Prisma, per `CLAUDE.md`) covering: a first request issues and sends; an immediate second request for the same identifier does not retire the first code.

</details>

---
**Why this matters:** The ADR frames "cannot be throttled without wrapping the handler" as a constraint, but the matcher is this repo's own and `src/server/services/whatsapp/VerificationService.ts` is in-repo precedent for exactly this control on exactly this concept. Same loop also amplifies a pre-existing oddity: Auth.js calls `callbacks.signIn` with `email: { verificationRequest: true }` at *send* time, so `config.ts:392` runs `acceptPendingInvitationsForUser` and stamps `lastLogin` before the address is ever proven — an attacker can force-accept a victim's pending workspace invitations without them signing in. Out of scope here, worth a follow-up ticket.

src/server/services/EmailService.ts
```diff
@@ -164,31 +172,23 @@ export async function sendMagicLinkEmail(
           <tr>
             <td style="padding: 0 32px;">
               <p style="margin: 0 0 24px; font-size: 15px; line-height: 1.6; color: #4b5563;">
-                Click the button below to securely access your account.
+                Enter this code on the sign-in page to access your account.
```
positonic, line 175: **[High] The sign-in email contains no URL at all, and the obvious recovery destroys the code**

The returning-user email now says "Enter this code on the sign-in page" and — verified across the whole template, body and footer — contains **zero** URLs. A user who reads it on their phone, or who closed the tab, has nothing to click and no domain to type. The natural recovery is to navigate to `/signin`, where the only visible action is *Send me a sign-in code* — which, since this PR retires an address's older codes (`config.ts:247`), silently kills the code they are looking at. They can loop that indefinitely.

The ADR's concern is that a scanner **spends a credential** link. A bare link to `/auth/verify-request` carries no token, so a scanner following it does nothing, and the page already handles the no-`sessionStorage` case by asking for the address. Adding it costs nothing the ADR was protecting against.

Suggested fix — add a non-credential link beneath the code, in both HTML and text bodies (and the same in `generateWelcomeEmailContent`'s `signInCode` branch):

```ts
const verifyUrl = `${getPublicBaseUrlFromEnv()}/auth/verify-request`;
// ...HTML, under the code block:
//   <p style="...">Enter it at <a href="${verifyUrl}">${verifyUrl}</a></p>
// ...text body:
//   `Enter it at: ${verifyUrl}`
```

<details>

In `src/server/services/EmailService.ts`, `sendSignInCodeEmail` (around line 175) renders a sign-in code with no URL anywhere in the email — not in the HTML body, not in the footer, not in the text body. That leaves a recipient reading the email on a different device with no way to reach the code-entry form, and the workaround (go to `/signin` and press "Send me a sign-in code") retires the code they already have, because `src/server/auth/config.ts:247` deletes an identifier's older tokens on each issue.

Add a plain, credential-free link to `/auth/verify-request` to both the HTML and text bodies of `sendSignInCodeEmail`, and to the `signInCode` branch of `generateWelcomeEmailContent` (whose only current link points at `/daily-plan`).

Constraints:
- Build the URL from `getPublicBaseUrlFromEnv()`, already imported in this file.
- The link must carry **no** token and no email address — that is the entire point of ADR-0056; a scanner following it must be a no-op.
- `src/app/(sidemenu)/auth/verify-request/page.tsx` already handles arriving with no `sessionStorage` by prompting for the address, so no page change should be needed — verify that.
- Keep the existing hardcoded email colours as they are; `EmailService.ts` is an allowlisted exception in `.husky/pre-commit`.
- Add a unit test asserting the rendered bodies contain the verify-request URL and contain no URL with a `token=` query parameter.

</details>

---
**Why this matters:** Verified by reading the full template (lines 140–225): the only text is "Didn't request this?" in the footer. No domain string at all. The welcome variant does carry URLs, but its only sign-in-adjacent one is `${appUrl}/daily-plan` — which lands a signed-out new user on a page that offers to send them a *second* code. This is the highest-frequency path in the feature (every returning email user, every send).

src/app/signin/SignInLandingPage.tsx
```diff
@@ -3,16 +3,54 @@
...
+ * else — including a failed send, which is `EmailSignInError` internally — is
+ * flattened to `Configuration` before it reaches the browser. So a failed send
+ * is caught at the call site instead, where we still know what happened.
+ */
+function signInErrorMessage(error: string | null): string | null {
+  switch (error) {
+    case null:
+      return null;
+    case "Verification":
+      return "That sign-in code is incorrect, has expired, or has already been used. Request a new one below.";
```
positonic, line 37: **[Medium] A mistyped code strands the user — `/signin` has no route back to the code form**

`pages.error` is `/signin`, so a wrong code lands here with `?error=Verification` and the copy "Request a new one below." But the original code is usually still valid for several more minutes, and `/signin` has no link to `/auth/verify-request` (grepped — the only three references are the two `router.push` calls and the `PublicForm` anchor). Eight hand-typed characters will be mistyped often, so the common case is: one typo → forced reissue → the code in the inbox is retired (`config.ts:247`) → wait for a new email. Two typos in a row and the user is convinced sign-in is broken.

Suggested fix — offer the way back, and stop telling them to burn a working code:

```suggestion
    case "Verification":
      return "That sign-in code didn't work — check it against the email, or request a new one below.";
```

…plus a link to the form, rendered next to the error:

```tsx
{searchParams.get("error") === "Verification" && (
  <p className="auth-sub">
    Still have your code? <Link href="/auth/verify-request">Enter it again</Link>.
  </p>
)}
```

<details>

In `src/app/signin/SignInLandingPage.tsx` at line 40, the `Verification` error case tells the user to request a new sign-in code. That is usually wrong: a mistyped code does not consume the stored `VerificationToken`, so the original is still valid for the rest of its 10-minute TTL — and requesting a new one now *retires* it (`src/server/auth/config.ts:247`).

Two changes:
1. Reword the `Verification` message so re-entering the existing code is the first suggestion and requesting a new one is the fallback.
2. Render a link to `/auth/verify-request` when `?error=Verification` is present, so the user can get back to the code-entry form. `sessionStorage` still holds their identifier at that point, so the form will not re-prompt for the address.

Constraints:
- Don't change the shape of `signInErrorMessage`; other cases must keep their current copy.
- Use `next/link` (already imported in this file), not `useRouter`.
- The link markup must live inside `.auth-surface` so `src/styles/auth-surface.css` applies.
- Add a Playwright assertion in `e2e/sign-in-code.spec.ts` (the "wrong code says so" test) that the link is present and navigates back to the code form.

</details>

---
**Why this matters:** The existing E2E test already drives exactly this path and stops at asserting the message, so extending it is cheap. Pairs with the email-link finding above: both are the same underlying gap — `/auth/verify-request` is reachable only by an in-tab `router.push` that happens once.

src/app/(public)/f/[slug]/_components/PublicForm.tsx
```diff
@@ -97,14 +102,22 @@ export function PublicForm({
...
         redirect: false,
         callbackUrl: '/',
       });
       if (res?.error) {
         setAccountError('Could not start account setup. Please try again.');
       } else {
+        // Only once the code is actually on its way, so a failed attempt
+        // doesn't prime the verify page for a code that was never issued.
+        window.sessionStorage.setItem(SIGN_IN_EMAIL_KEY, identifier);
+        window.sessionStorage.setItem(SIGN_IN_CALLBACK_KEY, '/');
```
Comment (positonic) [inline src/app/(public)/f/[slug]/_components/PublicForm.tsx:121]: **[Medium] Three copies of the email-sign-in handoff, already diverging**

`normalize → signIn("postmark", {redirect:false}) → check result.error → write both sessionStorage keys only on success` now exists three times: `SignInLandingPage.tsx:88-111`, `invite/[token]/page.tsx:319-341`, and here. The "only after the send succeeded" ordering is a security-relevant invariant that a fourth caller will get wrong — and `/f/[slug]` was already the caller that got missed once (per the PR description).

They have already drifted: the other two share `SEND_FAILED_MESSAGE` specifically "so their copy can't drift", while this one says "Could not start account setup. Please try again." (lines 115 and 124), and it hardcodes `SIGN_IN_CALLBACK_KEY` to `'/'` rather than a real destination.

Suggested fix — one client-safe helper in `~/lib/signInCode.ts`, alongside the keys it writes:

```ts
export async function startEmailSignIn(
  rawEmail: string,
  callbackUrl: string,
): Promise<{ ok: true } | { ok: false }> {
  const identifier = normalizeSignInEmail(rawEmail);
  const result = await signIn("postmark", { email: identifier, callbackUrl, redirect: false });
  if (result?.error) return { ok: false };
  window.sessionStorage.setItem(SIGN_IN_EMAIL_KEY, identifier);
  window.sessionStorage.setItem(SIGN_IN_CALLBACK_KEY, callbackUrl);
  return { ok: true };
}
```

<details>

Three components duplicate the email sign-in handoff: `src/app/signin/SignInLandingPage.tsx` (lines ~88-111), `src/app/invite/[token]/page.tsx` (lines ~319-341), and `src/app/(public)/f/[slug]/_components/PublicForm.tsx` (lines ~105-121). Each normalizes the address, calls `signIn("postmark", { redirect: false })`, checks `result?.error`, and writes `SIGN_IN_EMAIL_KEY` / `SIGN_IN_CALLBACK_KEY` to `sessionStorage` only after a successful send.

Extract that sequence into a single exported helper — `startEmailSignIn(rawEmail, callbackUrl)` — and have all three call it. Put it in `src/lib/signInCode.ts` next to the `sessionStorage` keys it writes, and move `SEND_FAILED_MESSAGE` there too (it currently lives in `SignInLandingPage.tsx` and is imported by the invite page).

Constraints:
- Preserve each surface's distinct behaviour after the call: `/signin` and `/invite` navigate via `router.push("/auth/verify-request")` and must keep their `navigating` flag so the button stays disabled through the navigation; `PublicForm` stays on the page and shows an inline anchor.
- `PublicForm` should use the shared `SEND_FAILED_MESSAGE` copy rather than its own two strings.
- Do not change the "write `sessionStorage` only on success" ordering — that is the invariant `e2e/sign-in-code.spec.ts` asserts.
- Keep `src/lib/signInCode.ts` importable from `src/server/auth/config.ts`, which `middleware.ts` pulls into the Edge bundle: keep the new helper in a separate `"use client"`-safe module if importing `next-auth/react` there would break that. Verify with `npm run check`.

</details>

---
**Why this matters:** `PublicForm` hardcoding `SIGN_IN_CALLBACK_KEY` to `'/'` means an applicant who signs in ends up on the marketing root rather than anywhere useful — arguably its own small bug. Extracting this also gives a single place to add the reissue-cooldown UI the throttling finding will need.

src/app/invite/[token]/page.tsx
```diff
@@ -24,6 +24,12 @@ import { signIn, signOut } from "next-auth/react";
 import { type FormEvent, useMemo, useState } from "react";
 import { api, type RouterOutputs } from "~/trpc/react";
 import { PRODUCT_NAME } from "~/lib/brand";
+import {
+  normalizeSignInEmail,
+  SIGN_IN_CALLBACK_KEY,
+  SIGN_IN_EMAIL_KEY,
+} from "~/lib/signInCode";
+import { SEND_FAILED_MESSAGE } from "~/app/signin/SignInLandingPage";
```
Comment (positonic) [inline src/app/invite/[token]/page.tsx]: **[Low] `SEND_FAILED_MESSAGE` is exported from a page component**

`~/app/signin/SignInLandingPage` is a `"use client"` page component that pulls in `~/trpc/react` and `~/styles/auth-surface.css`. Importing a string constant from it makes the invite route depend on the sign-in page's module graph, and makes a UI page the owner of shared auth copy. `~/lib/signInCode.ts` is already the shared home for `SIGN_IN_EMAIL_KEY`, `SIGN_IN_CALLBACK_KEY` and the TTL constants — the message belongs there too.

Suggested fix:
```suggestion
import { SEND_FAILED_MESSAGE } from "~/lib/signInCode";
```

<details>

`src/app/invite/[token]/page.tsx` line 32 imports `SEND_FAILED_MESSAGE` from `~/app/signin/SignInLandingPage`, a `"use client"` page component. Move the constant to `src/lib/signInCode.ts`, where the other shared sign-in constants already live, and update both importers (`SignInLandingPage.tsx` should import it rather than declare it).

Constraints:
- `src/lib/signInCode.ts` is imported by `src/server/auth/config.ts`, which `middleware.ts` pulls into the Edge runtime — a bare string constant is fine, but do not add any runtime imports to that module.
- Re-export from `SignInLandingPage.tsx` only if something outside this repo depends on it; otherwise update the import sites directly.
- Run `npm run check` afterwards.

</details>

---
**Why this matters:** Same fix as the duplication finding above; if that one is taken, this comes free.

positonic (commented):
Looks good.

PR #565, opened 2026-08-12
refactor: Fix the broken seams in the invite to first-session path

positonic:
## What

Six small, independent fixes that make the invite → signup → first session coherent. One commit per fix:

1. **"I'll explore on my own" is a real exit** — marks welcome complete (`user.completeWelcome`) and navigates to the default workspace home; rendered as a persistent quiet link in both the Chat and Checklist welcome views.
2. **No more "Invitation Already Used" dead end** — a logged-in invitee (or existing member) opening an accepted invite link is redirected straight to `/w/<slug>`; `getInvitationByToken` now reports viewer membership.
3. **Join activity for auto-accepted invites** — `acceptPendingInvitationsForUser` now emits the same `workspace_member joined` event as `workspace.acceptInvitation`, so invite signups appear in the team feed.
4. **Invite links use the shared public-URL helper** — `generateInviteUrl` / `generateTeamInviteUrl` no longer read `NEXTAUTH_URL` (localhost fallback); covered by unit tests.
5. **Invite email failures are reported** — all five invite / member-added send sites route their catch through `reportHandledError()` (Sentry + Bug ticket) instead of only console-logging; the mutations still succeed.
6. **Onboarding artifacts stay private** — `welcome.createGoal` / `createAction` land in the user's personal workspace when their default workspace is a shared team/organization workspace.

## Acceptance criteria

- [x] Clicking "I'll explore on my own" leaves /welcome, sets `welcomeCompletedAt`, and the affordance exists in both welcome views
- [x] A logged-in invitee opening an already-used invite link lands in the workspace without a manual click
- [x] Invites auto-accepted at signup produce a workspace-member joined activity event visible in the team feed
- [x] Invite emails link to the public app URL regardless of `NEXTAUTH_URL`
- [x] A failed invite/member-added email send is reported via `reportHandledError()` (Sentry + Bug ticket), not just logged
- [x] Onboarding goal/action never become visible to other members of a shared workspace

All six verified against a local dev server (real NextAuth sign-in for the auto-accept path, forced Postmark failure for the email path, per-member goal-listing check for workspace privacy).

---

Ships: slow.drift

[View in Exponential](https://www.exponential.im/w/syntrofi/tickets/cmsq4agbk000pju048d2och6q)

### DIFF
--- a/src/app/_components/welcome/useWelcomeSetup.ts
+++ b/src/app/_components/welcome/useWelcomeSetup.ts
@@ -242,11 +242,14 @@ export function useWelcomeSetup() {
         await completeWelcome.mutateAsync();
       } catch {
         // onError resets completedRef so a later attempt can retry; still
-        // leave — the workspace home doesn't gate on welcome completion.
+        // leave — neither destination below gates on welcome completion.
       }
     }
+    // Fallback is /today, NOT /home: /home bounces incomplete-welcome users
+    // back here during the 24h new-user window, which would turn a failed
+    // completeWelcome + unresolved workspace query into a loop.
     router.push(
-      defaultWorkspace?.slug ? `/w/${defaultWorkspace.slug}/home` : "/home",
+      defaultWorkspace?.slug ? `/w/${defaultWorkspace.slug}/home` : "/today",
     );
   }, [data, completeWelcome, defaultWorkspace, router]);
 
--- a/src/app/_components/welcome/useWelcomeSetup.ts
+++ b/src/app/_components/welcome/useWelcomeSetup.ts
@@ -59,6 +59,7 @@ export function useWelcomeSetup() {
   const router = useRouter();
 
   const { data, isLoading } = api.welcome.getSetup.useQuery();
+  const { data: defaultWorkspace } = api.workspace.getDefault.useQuery();
   const state = data?.state ?? null;
   const calendarConnected = data?.calendarConnected ?? false;
   // False while Google is verifying our calendar scopes and this user isn't an
@@ -229,6 +230,26 @@ export function useWelcomeSetup() {
     router.push("/today");
   }, [router]);
 
+  /**
+   * "I'll explore on my own" — a real exit, not a chat reply. Marks welcome
+   * complete (so /home stops bouncing back here for the 24h new-user window)
+   * and lands the user on their default workspace home.
+   */
+  const exploreOnOwn = useCallback(async () => {
+    if (!data?.welcomeCompletedAt && !completedRef.current) {
+      completedRef.current = true;
+      try {
+        await completeWelcome.mutateAsync();
+      } catch {
+        // onError resets completedRef so a later attempt can retry; still
+        // leave — the workspace home doesn't gate on welcome completion.
+      }
+    }
+    router.push(
+      defaultWorkspace?.slug ? `/w/${defaultWorkspace.slug}/home` : "/home",
+    );
+  }, [data, completeWelcome, defaultWorkspace, router]);
+
   const isAnswerPending =
     createGoal.isPending ||
     createAction.isPending ||
@@ -249,6 +270,7 @@ export function useWelcomeSetup() {
     answerStep,
     isAnswerPending,
     goToToday,
+    exploreOnOwn,
   };
 }
 
--- a/src/app/invite/[token]/page.tsx
+++ b/src/app/invite/[token]/page.tsx
@@ -70,12 +70,14 @@ export default function InviteAcceptPage() {
 
   // Invite signups are auto-accepted during user creation, so the invitee
   // returns here to an already-used invitation. That's success, not an error —
-  // send them straight into the workspace instead of dead-ending.
+  // send them straight into the workspace instead of dead-ending. Gated on
+  // actual membership (not isForCurrentUser): an invitee who was later
+  // removed from the workspace would only bounce off the access gate.
   const shouldAutoRedirect =
     !!invitation &&
     invitation.status === "accepted" &&
     invitation.isLoggedIn &&
-    (invitation.isForCurrentUser || invitation.isMember);
+    invitation.isMember;
   const workspaceSlug = invitation?.workspace.slug;
 
   useEffect(() => {
--- a/src/app/invite/[token]/page.tsx
+++ b/src/app/invite/[token]/page.tsx
@@ -21,7 +21,7 @@ import {
 } from "@tabler/icons-react";
 import { useParams, useRouter } from "next/navigation";
 import { signIn, signOut } from "next-auth/react";
-import { type FormEvent, useMemo, useState } from "react";
+import { type FormEvent, useEffect, useMemo, useState } from "react";
 import { api, type RouterOutputs } from "~/trpc/react";
 import { PRODUCT_NAME } from "~/lib/brand";
 import {
@@ -68,7 +68,23 @@ export default function InviteAcceptPage() {
     },
   });
 
-  if (isLoading) {
+  // Invite signups are auto-accepted during user creation, so the invitee
+  // returns here to an already-used invitation. That's success, not an error —
+  // send them straight into the workspace instead of dead-ending.
+  const shouldAutoRedirect =
+    !!invitation &&
+    invitation.status === "accepted" &&
+    invitation.isLoggedIn &&
+    (invitation.isForCurrentUser || invitation.isMember);
+  const workspaceSlug = invitation?.workspace.slug;
+
+  useEffect(() => {
+    if (shouldAutoRedirect && workspaceSlug) {
+      router.replace(`/w/${workspaceSlug}`);
+    }
+  }, [shouldAutoRedirect, workspaceSlug, router]);
+
+  if (isLoading || shouldAutoRedirect) {
     return (
       <Container size="sm" className="py-16">
         <Card className="bg-surface-secondary border-border-primary" withBorder>
--- a/src/server/api/routers/team.ts
+++ b/src/server/api/routers/team.ts
@@ -3,6 +3,7 @@ import { createTRPCRouter, protectedProcedure, publicProcedure } from "~/server/
 import { TRPCError } from "@trpc/server";
 import { generateSecureToken, generateTeamInviteUrl } from "~/server/utils/tokens";
 import { sendTeamInvitationEmail } from "~/server/services/EmailService";
+import { reportHandledError } from "~/lib/reportHandledError";
 
 export const teamRouter = createTRPCRouter({
   // Create a new team
@@ -322,6 +323,10 @@ export const teamRouter = createTRPCRouter({
           inviteUrl,
         }).catch((err: unknown) => {
           console.error("[team.addMember] Failed to send invitation email:", err);
+          reportHandledError(err, {
+            area: "team-invitation-email",
+            context: { teamId: input.teamId, recipientEmail: input.email },
+          });
         });
 
         return {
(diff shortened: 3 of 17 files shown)


src/server/api/routers/welcome.ts
```diff
@@ -65,8 +71,25 @@ async function resolveWorkspaceId(
...
     select: { defaultWorkspaceId: true },
   });
-  if (user?.defaultWorkspaceId) return user.defaultWorkspaceId;
+  const defaultWorkspace = user?.defaultWorkspaceId
+    ? await db.workspace.findUnique({
+        where: { id: user.defaultWorkspaceId },
+        select: { id: true, type: true },
+      })
+    : null;
+  if (defaultWorkspace?.type === "personal") return defaultWorkspace.id;
```
positonic: **[High] Onboarding goal/action still leaks into someone else's *personal* workspace**

`resolveWorkspaceId` trusts `type === "personal"` without checking ownership. Invites into personal-type workspaces are fully supported (the settings Members tab and `workspace.addMember` have no type gate), and signup auto-accept sets the invitee's `defaultWorkspaceId` to the inviter's workspace (`src/server/auth/config.ts`) — so a user invited into someone's Personal workspace gets their throwaway onboarding goal/action created in *that* workspace, visible to its owner. Same leak class this commit fixes, through a different door.

Suggested fix (every user gets an owned personal workspace at signup, so the fallback branch below resolves correctly):
```suggestion
  const defaultWorkspace = user?.defaultWorkspaceId
    ? await db.workspace.findUnique({
        where: { id: user.defaultWorkspaceId },
        select: { id: true, type: true, ownerId: true },
      })
    : null;
  if (defaultWorkspace?.type === "personal" && defaultWorkspace.ownerId === userId)
    return defaultWorkspace.id;
```

<details>

In `src/server/api/routers/welcome.ts`, the `resolveWorkspaceId` helper (around lines 66-99) returns the user's default workspace whenever its `type` is `"personal"` — without verifying the user owns it. A user can be invited into someone else's personal workspace (`workspace.addMember` has no workspace-type restriction), and the auth signup path sets `defaultWorkspaceId` to the first auto-accepted invite's workspace. In that case onboarding goals/actions created by `welcome.createGoal` / `welcome.createAction` land in the *inviter's* personal workspace, visible to them.

Fix: in the `db.workspace.findUnique` select, also select `ownerId`, and change the early return to require both `defaultWorkspace.type === "personal"` and `defaultWorkspace.ownerId === userId`. The existing fallback (`db.workspace.findFirst({ where: { ownerId: userId, type: "personal" } })`) then handles the invited-into-personal case correctly.

Constraints: don't change the function signature or the final fallback order; keep `npm run check` green; add an integration test (see `src/server/api/routers/__tests__/workspace.integration.test.ts` and `src/test/trpc-helpers.ts` `createTestCaller`) covering: default = own personal (used), default = team (falls back to owned personal), default = someone else's personal (falls back to owned personal).

</details>

src/server/api/routers/team.ts
```diff
@@ -322,6 +323,10 @@ export const teamRouter = createTRPCRouter({
           inviteUrl,
         }).catch((err: unknown) => {
           console.error("[team.addMember] Failed to send invitation email:", err);
+          reportHandledError(err, {
+            area: "team-invitation-email",
+            context: { teamId: input.teamId, recipientEmail: input.email },
+          });
```
positonic: **[Medium] Server-side `reportHandledError` never files the Bug ticket — only Sentry fires**

`reportHandledError` posts to `/api/client-errors` with a **relative URL** and no credentials. From a tRPC router that `fetch` rejects in Node ("Failed to parse URL"), and even with an absolute URL the route authenticates via the caller's session cookie, which a server-side fire-and-forget fetch doesn't carry. So for all five new server call sites (this one, `team.resendInvitation`, and the three in `workspace.ts`), the "Sentry + Bug ticket" behavior claimed in the PR description is half true: the ticket half is a silent no-op. Sentry capture still works.

Suggested fix: give the helper a server path — e.g. a `reportHandledErrorServer` in `src/server/` that calls the underlying `buildClientErrorBug`/`ingestSentryBug` services (`src/server/services/clientErrors/`, `src/server/services/sentry/`) directly instead of the HTTP round-trip — and use it from these routers.

<details>

`src/lib/reportHandledError.ts` is a client-oriented helper: besides `Sentry.captureException`, it does `fetch("/api/client-errors", ...)` with a relative URL and `keepalive`. This PR calls it from server-side tRPC routers (`src/server/api/routers/team.ts` lines ~326 and ~677, `src/server/api/routers/workspace.ts` lines ~667, ~755, ~1177). On the server the relative-URL fetch rejects (Node/undici requires an absolute URL) and is silently swallowed by the helper's own catch; additionally `/api/client-errors` (`src/app/api/client-errors/route.ts`) authenticates via the user's session cookie, which a server-originated fetch would not carry. Net effect: only the Sentry half fires from server code; no Bug ticket is ever filed.

Create a server-side variant (e.g. `src/server/utils/reportHandledErrorServer.ts`) that captures to Sentry the same way and files the bug by calling the existing server services directly — see `src/app/api/client-errors/route.ts` for how it uses `shouldFileClientError`, `buildClientErrorBug`, `buildClientErrorBody` and `ingestSentryBug`, including the `CLIENT_ERROR_BUGS` env gate. Then switch the five server call sites in team.ts / workspace.ts to it. Keep it best-effort: it must never throw or reject into the caller. Do not change the client helper's behavior. Note `src/server/api/routers/matrixServer.ts` line ~263 has the same pre-existing problem — fix it too if in scope. Keep `npm run check` and unit tests green.

</details>

src/app/invite/[token]/page.tsx
```diff
@@ -68,7 +68,23 @@ export default function InviteAcceptPage() {
...
 
-  if (isLoading) {
+  // Invite signups are auto-accepted during user creation, so the invitee
+  // returns here to an already-used invitation. That's success, not an error —
+  // send them straight into the workspace instead of dead-ending.
+  const shouldAutoRedirect =
+    !!invitation &&
+    invitation.status === "accepted" &&
+    invitation.isLoggedIn &&
+    (invitation.isForCurrentUser || invitation.isMember);
```
Comment (positonic) [inline src/app/invite/[token]/page.tsx]: **[Low] Redirecting on `isForCurrentUser` sends removed members into a workspace they can't access**

An invitee who accepted and was later removed from the workspace still satisfies `status === "accepted" && isForCurrentUser`, so their old invite email link now auto-redirects into `/w/<slug>`, where the default-deny gate bounces them with no explanation — worse than the old card, which at least said the invite was used. `isMember` alone covers both intended cases (auto-accepted signup; already joined), because membership is the thing the redirect actually needs.

```suggestion
    invitation.isMember;
```

<details>

In `src/app/invite/[token]/page.tsx` (around lines 74-78), `shouldAutoRedirect` is `invitation.status === "accepted" && invitation.isLoggedIn && (invitation.isForCurrentUser || invitation.isMember)`. The `isForCurrentUser` disjunct admits a user whose invitation was accepted but who has since been removed from the workspace — they get auto-redirected to a workspace the middleware will deny. Change the condition to require `invitation.isMember` (drop the `isForCurrentUser ||` part): the server-side `isMember` flag added in `workspace.getInvitationByToken` is computed per-request, so any auto-accepted invitee is already a member by the time this page loads. Keep the skeleton-while-redirecting render and the effect's dependency array consistent with the change. Keep `npm run check` green.

</details>

src/app/_components/welcome/useWelcomeSetup.ts
```diff
@@ -229,6 +230,26 @@ export function useWelcomeSetup() {
...
+      try {
+        await completeWelcome.mutateAsync();
+      } catch {
+        // onError resets completedRef so a later attempt can retry; still
+        // leave — the workspace home doesn't gate on welcome completion.
+      }
+    }
+    router.push(
+      defaultWorkspace?.slug ? `/w/${defaultWorkspace.slug}/home` : "/home",
+    );
```
positonic: **[Low] The `/home` fallback can bounce the user straight back to /welcome**

If `workspace.getDefault` hasn't resolved yet (or returns nothing) the exit routes to `/home` — which redirects incomplete-welcome users back to `/welcome` for the 24h new-user window (`src/app/(sidemenu)/home/page.tsx` → `resolveNewUserRedirect`). Combined with a failed `completeWelcome` (the swallowed catch above), "I'll explore on my own" becomes a loop back to the page the user just tried to leave — the comment's claim that the destination "doesn't gate on welcome completion" only holds for the `/w/<slug>/home` branch. `/today` doesn't gate and is already the flow's other exit.

```suggestion
      defaultWorkspace?.slug ? `/w/${defaultWorkspace.slug}/home` : "/today",
    );
```

<details>

In `src/app/_components/welcome/useWelcomeSetup.ts`, `exploreOnOwn` (around lines 238-251) falls back to `router.push("/home")` when `defaultWorkspace?.slug` is unavailable. `/home` (`src/app/(sidemenu)/home/page.tsx`) redirects users whose `welcomeCompletedAt` is null back to `/welcome` during the 24h new-user window via `resolveNewUserRedirect` — so if the `completeWelcome` mutation failed (its error is deliberately swallowed just above) and the `workspace.getDefault` query hasn't resolved, clicking "I'll explore on my own" loops the user back to /welcome. Change the fallback destination from "/home" to "/today", which does not gate on welcome completion and is already used by `goToToday` in the same hook. Don't change the primary `/w/<slug>/home` branch or the completeWelcome error handling. Keep `npm run check` green.

</details>

positonic (commented):
Solid PR overall — the six fixes are well-scoped, one commit each, and the event/URL unification matches the existing patterns. Requesting changes for one thing only: `resolveWorkspaceId` still leaks onboarding goals/actions into someone *else's* personal workspace (the exact bug class commit 6 fixes), because it checks `type === "personal"` but not ownership. One-line fix, inline. The rest (server-side `reportHandledError`'s Bug-ticket half being a no-op, two Low edge cases, missing tests for the privacy-critical path) is non-blocking.

---
**Adversarial review (/pr-review)** — 1 High, 2 Medium, 2 Low. Inline comments carry the point fixes; the general finding below has no in-diff anchor.

---

## [Medium] The privacy-critical `resolveWorkspaceId` and the auto-accept join event are untested

`src/server/api/routers/welcome.ts` `resolveWorkspaceId` is now the enforcement point for "onboarding artifacts never land in a shared workspace", and `acceptPendingInvitationsForUser` grew a guarded activity write — but the PR's only new tests cover the invite-URL helpers. The acceptance criterion is verified only by a one-off manual check; a future edit to either branch order silently reintroduces the leak.

The repo already has everything needed: `createTestCaller()` from `src/test/trpc-helpers.ts`, factories in `src/test/factories/`, and `workspace.integration.test.ts` as precedent. Worth covering at minimum:

- default workspace is the user's own personal workspace → used as-is
- default is a team/organization workspace → falls back to the owned personal workspace
- default is a personal workspace **owned by someone else** → falls back to the owned personal workspace (see the High inline comment)
- no personal workspace exists → previous resolution order preserved
- auto-accept emits exactly one `workspace_member/created` event for a new member, and none when the user was already a member

positonic (commented):
👍

src/server/api/routers/welcome.ts
positonic: ✅ Addressed in 956a4d48 — ownership check added (`type === "personal" && ownerId === userId`), plus unit tests covering every resolution branch including the invited-into-someone-else's-personal case.

src/server/api/routers/team.ts
positonic: ✅ Addressed in 704aea37 — new `reportHandledErrorServer` (`src/server/utils/reportHandledErrorServer.ts`) captures to Sentry identically and files the Bug ticket via the same `clientErrorBug`/`ingestSentryBug` services the route wraps, behind the same `CLIENT_ERROR_BUGS` gate. All five call sites switched, plus the two pre-existing ones in `matrixServer.ts`.

src/app/invite/[token]/page.tsx
Comment (positonic) [reply src/app/invite/[token]/page.tsx]: ✅ Addressed in a331a94b — dropped `isForCurrentUser`; the redirect now requires `isMember`.

src/app/_components/welcome/useWelcomeSetup.ts
positonic: ✅ Addressed in 0e516b4f — fallback destination is `/today`.

PR #573, opened 2026-08-13
feat: Mentions inbox reading the existing notification pipeline

positonic:
## What

The read side of the ADR-0045 notification pipeline — mention notifications were written end-to-end (`@[Name](userId)` parsing, recipient resolution, deduped Notification rows) but no UI ever read them back. This adds:

1. **Tracer: `notification.list` + Mentions inbox** — first reader of the `Notification` table: current user only, newest first, cursor-paginated, category filter. The Your work panel gains a Mentions section; each row deeplinks to the entity the mention lives on.
2. **Mark-as-read + mark-all, backed by a new `readAt` column** — the ticket pointed at the existing `markNotificationRead` mutation, but that targets the legacy `ScheduledNotification` table and the ADR-0045 table had no read state at all. Per refinement with James: `readAt DateTime?` on `Notification` (+ `(userId, readAt)` index, migration included), `notification.markRead` / `markAllRead` scoped to the current user via `updateMany`. Opening a mention marks it read.
3. **Unread count badge** — `notification.unreadCount` drives a badge on the section heading plus the Mark-all-read affordance; unread rows render bold with a dot.

## Migration

One additive migration: `20260812214409_add_notification_read_at` (nullable column + index — no backfill, no data risk). Note for the shared preview DB: per the develop-sync convention, sync develop after this merges or apply the migration to the preview DB.

## Acceptance criteria

- [x] Being @-mentioned in any comment type surfaces in the mentions inbox
- [x] Read/unread state persists and the unread count matches
- [x] Clicking a mention navigates to the entity it lives on
- [x] Only the current user's notifications are ever returned (every query/mutation filters on `ctx.session.user.id`)

Verified end-to-end in a local dev server: a real action-comment mention → Notification row → inbox row (bold + badge) → click navigates to the action and marks it read → mark-all clears the badge → state persists across reload.

---

Ships: rainy.gecko

[View in Exponential](https://www.exponential.im/w/syntrofi/tickets/cmsq4b7o20005in04l7mwx4qd)

### DIFF
--- a/src/server/api/routers/notification.ts
+++ b/src/server/api/routers/notification.ts
@@ -72,7 +72,7 @@ export const notificationRouter = createTRPCRouter({
           // A future scheduledFor means the notification hasn't fired yet.
           OR: [{ scheduledFor: null }, { scheduledFor: { lte: new Date() } }],
         },
-        orderBy: { createdAt: "desc" },
+        orderBy: [{ createdAt: "desc" }, { id: "desc" }],
         take: limit + 1,
         ...(input?.cursor ? { cursor: { id: input.cursor }, skip: 1 } : {}),
         select: {
@@ -88,7 +88,8 @@ export const notificationRouter = createTRPCRouter({
 
       let nextCursor: string | undefined;
       if (rows.length > limit) {
-        nextCursor = rows.pop()!.id;
+        rows.pop();
+        nextCursor = rows[rows.length - 1]?.id;
       }
       return { notifications: rows, nextCursor };
     }),
--- a/src/server/api/routers/notification.ts
+++ b/src/server/api/routers/notification.ts
@@ -82,6 +82,7 @@ export const notificationRouter = createTRPCRouter({
           message: true,
           deeplink: true,
           createdAt: true,
+          readAt: true,
         },
       });
 
@@ -90,6 +91,54 @@ export const notificationRouter = createTRPCRouter({
         nextCursor = rows.pop()!.id;
       }
       return { notifications: rows, nextCursor };
+    }),
+
+  /**
+   * Mark one Notification read. Scoped to the current user via updateMany so
+   * a foreign id is a silent no-op rather than an oracle; idempotent (an
+   * already-read row keeps its original readAt).
+   */
+  markRead: protectedProcedure
+    .input(z.object({ notificationId: z.string() }))
+    .mutation(async ({ ctx, input }) => {
+      await ctx.db.notification.updateMany({
+        where: {
+          id: input.notificationId,
+          userId: ctx.session.user.id,
+          readAt: null,
+        },
+        data: { readAt: new Date() },
+      });
+      return { success: true };
+    }),
+
+  /** Mark every unread Notification read, optionally within one category. */
+  markAllRead: protectedProcedure
+    .input(z.object({ category: z.string().optional() }).optional())
+    .mutation(async ({ ctx, input }) => {
+      const result = await ctx.db.notification.updateMany({
+        where: {
+          userId: ctx.session.user.id,
+          readAt: null,
+          ...(input?.category ? { category: input.category } : {}),
+        },
+        data: { readAt: new Date() },
+      });
+      return { success: true, count: result.count };
+    }),
+
+  /** Unread Notification count for the badge, optionally per category. */
+  unreadCount: protectedProcedure
+    .input(z.object({ category: z.string().optional() }).optional())
+    .query(async ({ ctx, input }) => {
+      return ctx.db.notification.count({
+        where: {
+          userId: ctx.session.user.id,
+          readAt: null,
+          ...(input?.category ? { category: input.category } : {}),
+          OR: [{ scheduledFor: null }, { scheduledFor: { lte: new Date() } }],
+        },
+      });
     }),
 
   // Get user notification preferences
--- a/src/server/api/routers/notification.ts
+++ b/src/server/api/routers/notification.ts
@@ -62,7 +62,7 @@ export const notificationRouter = createTRPCRouter({
     .input(
       z
         .object({
-          category: z.string().optional(),
+          category: z.enum(CATEGORY_LIST).optional(),
           limit: z.number().int().min(1).max(50).default(20),
           cursor: z.string().optional(),
         })
@@ -119,7 +119,7 @@ export const notificationRouter = createTRPCRouter({
 
   /** Mark every unread Notification read, optionally within one category. */
   markAllRead: protectedProcedure
-    .input(z.object({ category: z.string().optional() }).optional())
+    .input(z.object({ category: z.enum(CATEGORY_LIST).optional() }).optional())
     .mutation(async ({ ctx, input }) => {
       const result = await ctx.db.notification.updateMany({
         where: {
@@ -135,7 +135,7 @@ export const notificationRouter = createTRPCRouter({
 
   /** Unread Notification count for the badge, optionally per category. */
   unreadCount: protectedProcedure
-    .input(z.object({ category: z.string().optional() }).optional())
+    .input(z.object({ category: z.enum(CATEGORY_LIST).optional() }).optional())
     .query(async ({ ctx, input }) => {
       return ctx.db.notification.count({
         where: {
--- a/src/server/api/routers/notification.ts
+++ b/src/server/api/routers/notification.ts
@@ -46,6 +46,11 @@ async function resolveChannelAvailability(db: PrismaClient, userId: string) {
   };
 }
 
+/** Visibility window shared by list/unreadCount/markAllRead — a future scheduledFor means the notification hasn't fired yet. */
+const firedWindow = () => ({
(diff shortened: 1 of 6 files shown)


src/server/api/routers/notification.ts
```diff
@@ -47,6 +47,100 @@ async function resolveChannelAvailability(db: PrismaClient, userId: string) {
...
+          deeplink: true,
+          createdAt: true,
+          readAt: true,
+        },
+      });
+
+      let nextCursor: string | undefined;
+      if (rows.length > limit) {
+        nextCursor = rows.pop()!.id;
+      }
```
positonic, line 97: **[High] Cursor pagination silently drops one notification at every page boundary**

The query mixes two incompatible idioms: `cursor: { id }, skip: 1` (Prisma-docs style, where `nextCursor` must be the **last returned** row) with `nextCursor = rows.pop()!.id` (tRPC-docs style, which requires **no** `skip`). Prisma's cursor is inclusive, so with `skip: 1` the popped row — the first row of the next page — is skipped when the client passes it back: page 1 returns rows 1–5 with `nextCursor` = row 6, page 2 starts at row 7. Row 6 is never shown. The inbox UI doesn't paginate yet (limit 5), but the first consumer that does will silently lose one notification per page — the worst kind of inbox bug to debug later.

Separately, `orderBy: { createdAt: "desc" }` has no unique tie-breaker, so rows sharing a `createdAt` millisecond make cursor positioning non-deterministic (dupes/skips across pages).

Suggested fix (keeps `skip: 1`, matches the already-correct pattern in `crmApi.ts:260-266`):
```suggestion
      let nextCursor: string | undefined;
      if (rows.length > limit) {
        rows.pop();
        nextCursor = rows[rows.length - 1]?.id;
      }
```
…and change line 75 to `orderBy: [{ createdAt: "desc" }, { id: "desc" }],` for a deterministic order.

<details>

In `src/server/api/routers/notification.ts`, the `list` procedure (roughly lines 66–93) has an off-by-one in its cursor pagination. It queries with `take: limit + 1` and `...(input?.cursor ? { cursor: { id: input.cursor }, skip: 1 } : {})`, then sets `nextCursor = rows.pop()!.id`. Because Prisma's cursor is inclusive and `skip: 1` skips the cursor row itself, setting `nextCursor` to the popped (limit+1)-th row means that row — the first row of the next page — is skipped on the next request. Fix by keeping `skip: 1` but setting `nextCursor` to the last row actually returned: `rows.pop(); nextCursor = rows[rows.length - 1]?.id;` (this matches the correct existing pattern in `src/server/api/routers/crmApi.ts` around lines 260–266). Also change `orderBy: { createdAt: "desc" }` to `orderBy: [{ createdAt: "desc" }, { id: "desc" }]` so rows with identical `createdAt` timestamps paginate deterministically.

Constraints:
- Don't change the procedure's input/output shape (`{ notifications, nextCursor }`).
- Keep the `userId` scoping and the `scheduledFor` visibility filter exactly as they are.
- Add a test (mocked Prisma via `createMockCaller` from `src/test/trpc-helpers.ts`, or an integration test per repo convention) that paginates two pages of 3 over 7 rows and asserts the pages are contiguous with no dropped or duplicated ids.
- Run `npm run check` and keep existing tests green.

</details>

src/server/api/routers/notification.ts
```diff
@@ -47,6 +47,100 @@ async function resolveChannelAvailability(db: PrismaClient, userId: string) {
...
+  /** Mark every unread Notification read, optionally within one category. */
+  markAllRead: protectedProcedure
+    .input(z.object({ category: z.string().optional() }).optional())
+    .mutation(async ({ ctx, input }) => {
+      const result = await ctx.db.notification.updateMany({
+        where: {
+          userId: ctx.session.user.id,
+          readAt: null,
+          ...(input?.category ? { category: input.category } : {}),
+        },
```
positonic, line 130: **[Medium] `markAllRead` marks not-yet-fired scheduled notifications as read**

`list` and `unreadCount` both exclude rows with a future `scheduledFor` ("hasn't fired yet"), but `markAllRead` doesn't. A user clicking "Mark all read" will pre-read notifications they have never been shown — when the cron later fires a scheduled row (e.g. a due-date reminder or digest), it arrives already-read and never surfaces as unread. The mention-scoped UI dodges this today (mentions emit with `scheduledFor: null`), but the procedure is generic and the un-scoped call (`markAllRead({})`) hits it immediately.

Suggested fix:
```suggestion
        where: {
          userId: ctx.session.user.id,
          readAt: null,
          ...(input?.category ? { category: input.category } : {}),
          // Same visibility window as `list`/`unreadCount` — never pre-read
          // a notification that hasn't fired yet.
          OR: [{ scheduledFor: null }, { scheduledFor: { lte: new Date() } }],
        },
```

<details>

In `src/server/api/routers/notification.ts`, the `markAllRead` mutation (roughly lines 116–128) runs `updateMany` with `where: { userId, readAt: null, ...category }` — but unlike the sibling `list` and `unreadCount` procedures in the same file, it omits the visibility filter `OR: [{ scheduledFor: null }, { scheduledFor: { lte: new Date() } }]`. This means "Mark all read" also marks notifications whose `scheduledFor` is in the future — rows the user has never seen — so when the cron worker later fires them they arrive pre-read. Add the same `OR` visibility clause to `markAllRead`'s `where` so all three procedures share one visibility window.

Constraints:
- Don't change the mutation's input or return shape (`{ success, count }`).
- Keep the `userId` scoping and `readAt: null` guard exactly as they are.
- Consider extracting the shared visibility clause into a small local helper (e.g. `const firedWindow = () => ({ OR: [{ scheduledFor: null }, { scheduledFor: { lte: new Date() } }] })`) used by `list`, `unreadCount`, and `markAllRead`, so the three can't drift again — match the file's existing style.
- Run `npm run check` and keep existing tests green.

</details>

src/server/api/routers/notification.ts
```diff
@@ -47,6 +47,100 @@ async function resolveChannelAvailability(db: PrismaClient, userId: string) {
...
+   * List the current user's Notification rows — the read side of the
+   * ADR-0045 pipeline (which until this query only ever wrote them).
+   * Newest first, cursor-paginated, optionally narrowed to one category
+   * (e.g. "mention" for the home-panel inbox).
+   */
+  list: protectedProcedure
+    .input(
+      z
+        .object({
+          category: z.string().optional(),
```
positonic: **[Low] `category` inputs are free-form strings while the enum exists in the same file**

`list`, `markAllRead`, and `unreadCount` all take `category: z.string().optional()`, but this file already imports `CATEGORY_LIST` and uses `z.enum(CATEGORY_LIST)` in `setChannelPreference`. With a free string, a typo'd category (client hardcodes `'mention'`) silently returns an empty list / zero count instead of failing loudly. Tightening to the enum makes the vocabulary checkable end-to-end.

Suggested fix:
```suggestion
          category: z.enum(CATEGORY_LIST).optional(),
```
(same change in `markAllRead` and `unreadCount`; optionally have `YourWorkPanel` import `NOTIFICATION_CATEGORIES` from `~/server/services/notifications/emit/constants` instead of the string literal — the module has no server-only imports.)

<details>

In `src/server/api/routers/notification.ts`, the new `list` (line ~60), `markAllRead` (~117), and `unreadCount` (~132) procedures accept `category: z.string().optional()`. The file already imports `CATEGORY_LIST` from `~/server/services/notifications/emit/constants` and uses `z.enum(CATEGORY_LIST)` in `setChannelPreference`. Change the three new procedures to `category: z.enum(CATEGORY_LIST).optional()` so an invalid category is a validation error instead of a silent empty result. Then check `src/app/_components/home/YourWorkPanel.tsx` (the `'mention'` literals around lines 102–104 and 273) still typechecks — optionally replace the literals with `NOTIFICATION_CATEGORIES.MENTION` imported from the same constants module (it has no server-only dependencies, so it's safe in a client component).

Constraints:
- Don't rename or restructure the procedures.
- Run `npm run check`; the tRPC client types must still compile everywhere `notification.list` / `unreadCount` / `markAllRead` are called.

</details>

positonic (commented):
👍

src/server/api/routers/notification.ts
positonic: ✅ Addressed in 96e8f764 — nextCursor is now the last returned row (crmApi.ts pattern), orderBy tie-broken with id desc, and a new pagination test walks 7 rows in pages of 3 asserting contiguity.

src/server/api/routers/notification.ts
positonic: ✅ Addressed in 079aa96b — a shared `firedWindow()` helper now gates `list`, `unreadCount`, and `markAllRead` so the three can't drift.

src/server/api/routers/notification.ts
positonic: ✅ Addressed in b6387ec8 — all three inputs use `z.enum(CATEGORY_LIST)`; YourWorkPanel passes `NOTIFICATION_CATEGORIES.MENTION` (constants module verified client-safe).

PR #594, opened 2026-08-16
feat: V1 ICS calendar feeds end-to-end

positonic:
## What

Per-user **ICS calendar feeds** as a first-class calendar source (V1 of the Calendar feeds & workspace scheduling feature): a `CalendarFeed` model with the subscription URL encrypted at rest and SSRF-guarded, a new `CalendarEvent` persistence table (the app's first event persistence layer) synced by a 15-minute cron with a manual Refresh-now, full recurrence expansion (RRULE/EXDATE/RECURRENCE-ID, all-day + TZID/DST), rendering on /calendar alongside Google/Microsoft, propagation through the shared multi-calendar read paths (Today, home snapshot, agent tools), and a timezone checkpoint (`User.timezone`) whenever a calendar source is connected.

**Migration** (`20260816133134_add_calendar_feeds_events_user_timezone`): `CalendarFeed` + `CalendarEvent` + `User.timezone` — applies on merge via the production migrate step.

Decisions: ADR-0057 (feeds are deliberately NOT ConnectedAccounts), CONTEXT.md "Calendar feed" term.

Verified end-to-end against a scratch Postgres + seeded fixture user: add feed → inline sync → events on /calendar week grid + Today page, timezone prompt prefilled from X-WR-TIMEZONE, feed management in sidebar + settings, cron 401s without CRON_SECRET.

## Acceptance criteria

- [x] All 7 V1 requirement rows on the feature are met (encrypted CalendarFeed on submit; multiple feeds + disconnect; timezone prompt; ≤30-min cron persistence; Refresh now; failure retains events; shared-path merge)
- [x] providerSchema enum untouched; ICS enters only read-path merges and response payload types
- [x] SSRF guard: https-only, public hosts only, size cap + timeout on every feed fetch

---

Ships: bold.drift


src/server/services/calendar/CalendarSyncService.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  let parsed: ParsedIcsFeed;
+  const syncedAt = new Date();
+  try {
+    const icsText = await fetchIcsText(url, resolveHost);
+    parsed = parseIcsFeed(icsText, getSyncWindow(syncedAt));
+  } catch (error) {
+    const message = error instanceof Error ? error.message : "Unknown sync error";
+    return recordFailure(message);
+  }
```
positonic, line 352: **[High] A feed that 200s with non-ICS content silently wipes all events and reports "ok"**

`parseIcsFeed` returns `events: []` for arbitrary non-calendar text without throwing (pinned by the "returns no events for text that is not an ICS file" test), so `syncFeed` runs the delete-and-replace with zero rows and sets `syncStatus: "ok"`. Expired/revoked Outlook published-calendar links do exactly this in practice: they respond 200 with an HTML sign-in page. That deletes every event with no error recorded — the opposite of the documented contract ("a failed fetch retains the last good events and records the error on the feed", CONTEXT.md / ADR-0057).

Suggested fix: treat a body with no VCALENDAR as a sync failure before the transaction —

```ts
if (!/^BEGIN:VCALENDAR/m.test(icsText)) {
  return recordFailure("The URL did not return an ICS calendar (got HTML or other content).");
}
```

A genuinely emptied calendar still parses as a valid, empty VCALENDAR, so legitimate zero-event syncs are unaffected.

<details>

In `src/server/services/calendar/CalendarSyncService.ts`, `syncFeed` calls `parseIcsFeed(icsText, …)` (~line 272) and then unconditionally runs a delete-and-replace transaction. `parseIcsFeed` returns an empty `events` array for non-ICS text (e.g. an HTML sign-in page served with HTTP 200 by an expired Outlook published-calendar link) instead of throwing, so all previously synced `CalendarEvent` rows for the feed are deleted and `syncStatus` is set to `"ok"`. Required behavior: a response that is not an ICS calendar must be treated as a sync failure (call `recordFailure`), retaining the previous events, per the contract in CONTEXT.md. Add a guard that the fetched text contains a `BEGIN:VCALENDAR` component (or that the parse yields one) before the transaction. Constraints: keep `parseIcsFeed` pure and its existing tests green (including the "not an ICS file" test — you may adjust it to assert the new failure behavior at the `syncFeed` level); add a regression test where the fetch resolves with HTML and the previous events survive with `syncStatus: "error"`.

</details>

src/server/services/calendar/CalendarSyncService.ts
```diff
@@ -0,0 +1,358 @@
...
+  const feed = await db.calendarFeed.findUnique({ where: { id: feedId } });
+  if (!feed) return { ok: false, error: "Feed not found" };
+
+  const recordFailure = async (message: string): Promise<SyncFeedResult> => {
+    await db.calendarFeed.update({
+      where: { id: feedId },
+      data: { syncStatus: "error", lastSyncError: message.slice(0, 500) },
+    });
+    return { ok: false, error: message };
+  };
```
positonic: **[High] Failing feeds never advance the sweep queue — starvation + endless retry**

`recordFailure` updates `syncStatus`/`lastSyncError` but never a timestamp, and `runCalendarSync` orders by `lastSyncedAt asc, nulls first` with `take: 50`. A feed that always fails keeps `lastSyncedAt` null (or stale) forever, so it sorts to the head of **every** 15-minute sweep and is re-fetched in perpetuity with no backoff. Once broken feeds ≥ 50 — or far fewer if they're slow-timeout feeds eating the 300 s `maxDuration` (~20 × 15 s timeouts kills the function mid-sweep) — healthy feeds are never reached again, permanently and silently.

Suggested fix: add a `lastSyncAttemptAt` column, stamp it at the start of every `syncFeed` (success or failure), and order the sweep by it. Consider skipping feeds in `error` state until a backoff interval has passed.

<details>

In `src/server/services/calendar/CalendarSyncService.ts`: `recordFailure` (~line 253) records only `syncStatus: "error"` and `lastSyncError`; `runCalendarSync` (~line 337) selects enabled feeds ordered by `{ lastSyncedAt: { sort: "asc", nulls: "first" } }` with `take: 50`. Because failures never advance `lastSyncedAt`, permanently failing feeds monopolize the head of every cron batch: with ≥50 broken enabled feeds, healthy feeds are never synced again. Fix: add a `lastSyncAttemptAt DateTime?` to the `CalendarFeed` Prisma model (schema change → ask the user to run `npx prisma migrate dev`, never run it yourself), stamp it in `syncFeed` on every attempt, and change the sweep's `orderBy` to it. Also use it (instead of `lastSyncedAt`) for the `refreshMyFeeds` rate limit in `src/server/api/routers/calendar.ts` (~line 299), which the same gap currently disables for any user with a failing feed. Constraints: keep the "failures retain prior events" behavior; update the router tests in `src/server/api/routers/__tests__/calendarFeeds.test.ts` that currently assert rate-limit behavior via `lastSyncedAt`.

</details>

src/server/services/calendar/CalendarSyncService.ts
```diff
@@ -0,0 +1,358 @@
...
+export function parseIcsFeed(
+  icsText: string,
+  window: { from: Date; to: Date },
+): ParsedIcsFeed {
+  const parsed = ical.sync.parseICS(icsText);
+
+  const calendarName = parsed.vcalendar?.["WR-CALNAME"];
+  const calendarTimezone = parsed.vcalendar?.["WR-TIMEZONE"];
+
+  const events: NormalizedEvent[] = [];
```
positonic, line 101: **[High] Aggregate recurrence expansion is uncapped — a crafted feed can OOM the sync function**

rrule-temporal caps a *single* rule at 10 000 iterations (it throws), but the cap is per-rule. A 5 MB ICS file fits tens of thousands of ~150-byte VEVENTs each carrying `RRULE:FREQ=HOURLY` (≈1 500 instances per event over the 9-week window, under the per-rule cap) — tens of millions of `NormalizedEvent`s accumulated in one array, then attempted as one `createMany`. The function dies on memory/time. It's user-triggerable via `addFeed`'s inline sync, and because an OOM'd sync never reaches `recordFailure`, the feed stays at the head of the cron queue and crashes every subsequent sweep for **all** users.

Suggested fix: cap total expanded events per feed and fail the sync when exceeded —

```ts
const MAX_EVENTS_PER_FEED = 10_000;
// in parseIcsFeed, after each push:
if (events.length > MAX_EVENTS_PER_FEED) {
  throw new FeedFetchError("Feed expands to too many events.");
}
```

Also wrap the per-event `expandRecurringEvent` call in try/catch so one malformed RRULE (which *does* throw at the 10k per-rule cap) skips that event instead of failing the entire feed.

<details>

In `src/server/services/calendar/CalendarSyncService.ts`, `parseIcsFeed` (~lines 83–157) expands every recurring VEVENT via `ical.expandRecurringEvent` and accumulates all instances into one `events` array with no aggregate cap; `syncFeed` then inserts them all in one `createMany`. node-ical's rrule backend (rrule-temporal) throws past 10 000 iterations for a single rule, but a feed with many separate RRULE VEVENTs (e.g. thousands of `FREQ=HOURLY` events in a 5 MB file) can still expand to tens of millions of instances in aggregate and OOM the serverless function. Add: (1) a total-events cap per feed (~10 000) that aborts the parse with a thrown error so `syncFeed` records it as a sync failure and prior events are retained; (2) a per-event try/catch around the expansion so a single malformed RRULE (per-rule iteration-cap throw) is skipped rather than failing the whole feed. Constraints: keep `parseIcsFeed` pure; add unit tests for both behaviors (aggregate cap exceeded → throws; one bad RRULE among good events → good events still returned); keep all existing fixture tests green.

</details>

prisma/schema.prisma
```diff
@@ -2011,6 +2017,60 @@ model CalendarPreference {
...
+  title              String?
+  location           String?
+  startsAt           DateTime
+  endsAt             DateTime
+  isAllDay           Boolean       @default(false)
+  syncedAt           DateTime
+  user               User          @relation(fields: [userId], references: [id], onDelete: Cascade)
+  calendarFeed       CalendarFeed? @relation(fields: [calendarFeedId], references: [id], onDelete: Cascade)
+
+  @@unique([userId, sourceType, externalId, startsAt])
```
positonic: **[Medium] Cross-feed UID collision: unique key isn't feed-scoped, shared events silently drop**

`@@unique([userId, sourceType, externalId, startsAt])` spans *all* of a user's feeds, and `syncFeed` deletes only its own feed's rows before `createMany({ skipDuplicates: true })`. If the same event (same UID + start) appears in two feeds — the same calendar added twice, or two published views sharing a meeting — the row belongs to whichever feed synced first, and the other feed's copy is silently discarded. Worse: disable the "owning" feed and the event vanishes from the calendar entirely (the read path filters `calendarFeed.isEnabled`, and the disabled feed's stale row blocks the enabled feed from ever inserting its copy).

Suggested fix: scope ICS uniqueness by feed:

```prisma
@@unique([calendarFeedId, externalId, startsAt])
```

(V2's provider rows can get their own key shape on `connectedAccountId` when they land.) This also makes `eventCount` honest.

<details>

In `prisma/schema.prisma`, the `CalendarEvent` model's unique constraint is `@@unique([userId, sourceType, externalId, startsAt])` (line ~2069). Combined with `syncFeed` in `src/server/services/calendar/CalendarSyncService.ts` deleting only `where: { calendarFeedId: feedId }` and inserting with `skipDuplicates: true`, an event with the same iCal UID and start appearing in two of a user's feeds is silently dropped from the second feed, and disabling the first feed makes the event invisible everywhere (read path filters on `calendarFeed.isEnabled`). Change the constraint to be feed-scoped: `@@unique([calendarFeedId, externalId, startsAt])`. This is a schema change: modify schema.prisma and ASK THE USER to run `npx prisma migrate dev --name scope_calendar_event_uniqueness_by_feed` — never run migrations yourself in this repo. Constraints: keep the `@@index([userId, startsAt, endsAt])` read index; check the integration test `src/server/services/calendar/__tests__/CalendarSyncService.integration.test.ts` for assertions on the old constraint and update them; add a test covering two feeds containing the same UID+start where both feeds' events remain independently visible.

</details>

src/server/api/routers/calendar.ts
```diff
@@ -189,6 +193,142 @@ async function resolveAccount(
...
+        message: "Feeds were just refreshed — try again in a minute.",
+      });
+    }
+
+    let refreshed = 0;
+    for (const feed of feeds) {
+      const result = await syncFeed(ctx.db as PrismaClient, feed.id);
+      if (result.ok) refreshed += 1;
+    }
+    return { refreshed };
```
positonic, line 341: **[Medium] Rate limit disengages on any failing feed; inline loop unbounded vs 60 s route budget**

The 1/min guard fires only when *every* enabled feed is fresh. Since failing feeds never update `lastSyncedAt`, one broken feed makes `allFresh` false forever — each click then re-fetches **all** enabled feeds, sequentially, with no batch cap. Worst case per feed is ~60 s (redirect chain × 15 s timeouts) while the tRPC route has `maxDuration = 60` — a user with a few slow feeds gets a 504, and a spamming user gets unlimited outbound fetch amplification. The same budget applies to `addFeed`'s inline first sync: a slow feed 504s the mutation *after* the row was created, and a retry creates a duplicate feed.

Suggested fix: rate-limit on the most recent sync *attempt* across feeds (see the `lastSyncAttemptAt` suggestion on `CalendarSyncService.ts`), skip feeds that are already fresh instead of re-syncing them, and cap the number of feeds per invocation.

<details>

In `src/server/api/routers/calendar.ts`, `refreshMyFeeds` (~lines 290–316) rate-limits by checking that every enabled feed's `lastSyncedAt` is within the last minute, then re-syncs ALL enabled feeds sequentially. Problems: (1) feeds whose syncs fail never update `lastSyncedAt` (see `recordFailure` in `CalendarSyncService.ts`), so one failing feed permanently disables the rate limit; (2) the loop has no cap and each `syncFeed` can take up to ~60 s, while `src/app/api/trpc/[trpc]/route.ts` sets `maxDuration = 60`. Fix by: rate-limiting on a per-attempt timestamp (`lastSyncAttemptAt` — coordinate with the schema change suggested in the CalendarSyncService review comment), only re-syncing feeds that aren't already fresh, and capping feeds per invocation (e.g. 10). Also consider a shorter fetch timeout for the inline `addFeed` first sync so the mutation can't outlive the route budget. Constraints: keep the existing TOO_MANY_REQUESTS error shape (the calendar page swallows it); update the router tests in `src/server/api/routers/__tests__/calendarFeeds.test.ts`.

</details>

src/server/utils/privateAddress.ts
```diff
@@ -0,0 +1,59 @@
...
+function isPrivateIPv6(address: string): boolean {
+  const normalized = address.toLowerCase().replace(/^\[|\]$/g, "");
+  if (normalized === "::1" || normalized === "::") return true;
+  if (normalized.startsWith("fe80")) return true; // link-local
+  if (/^f[cd]/.test(normalized)) return true; // unique local
+  // IPv4-mapped (::ffff:169.254.169.254) — judge the embedded address.
+  const mapped = /^::ffff:(\d+\.\d+\.\d+\.\d+)$/.exec(normalized);
+  if (mapped) return isPrivateIPv4(mapped[1]!);
+  return false;
+}
```
positonic, line 118: **[Medium] SSRF guard misses non-dotted IPv4-mapped IPv6 and NAT64 forms**

`isPrivateIPv6` only recognizes the dotted-quad mapped form (`/^::ffff:(\d+\.\d+\.\d+\.\d+)$/`). Verified in Node: `::ffff:7f00:1` (hex form of 127.0.0.1), `0:0:0:0:0:ffff:127.0.0.1` (uncompressed), `::ffff:a9fe:a9fe` (169.254.169.254 — cloud metadata), and NAT64 `64:ff9b::7f00:1` all pass as public. A feed URL like `https://[::ffff:7f00:1]/` clears the guard and `fetch` connects to loopback. TLS certificate validation blunts most https exploitation, but the guard's contract is broken, error-message differences make it a port-probe oracle, and the gap now also applies to the Matrix homeserver guard sharing this module.

Suggested fix: canonicalize before judging — expand the address to its full form, extract any embedded IPv4 from `::ffff:0:0/96` and `64:ff9b::/96`, and judge the embedded address with `isPrivateIPv4`.

<details>

In `src/server/utils/privateAddress.ts`, `isPrivateIPv6` (~lines 43–52) detects IPv4-mapped addresses only in dotted form (`::ffff:127.0.0.1`) via regex. Hex-form mapped addresses (`::ffff:7f00:1`), uncompressed forms (`0:0:0:0:0:ffff:127.0.0.1`), and NAT64 (`64:ff9b::/96`) evade it, so `https://[::ffff:7f00:1]/` passes `assertSafeFeedUrl` and the server connects to loopback. Fix by normalizing the IPv6 address to full hextet form first (pure string manipulation is fine: split on `::`, pad to 8 groups, convert any trailing dotted quad), then: treat `::ffff:xxxx:xxxx` and `64:ff9b::` prefixes as carrying an embedded IPv4 and judge it with `isPrivateIPv4`; keep the existing loopback/link-local/ULA checks working on the normalized form. Constraints: this module is shared by the Matrix homeserver guard (`src/server/services/matrix/homeserverUrl.ts`) and the calendar feed guard — don't change the exported API (`isPrivateAddress`, `defaultResolveHost`, `ResolveHost`); add unit tests in `src/server/services/calendar/__tests__/feedUrlGuard.test.ts` for each bypass form listed above; no new dependencies unless one is already in package.json.

</details>

src/app/api/cron/sync-calendars/route.ts
```diff
@@ -0,0 +1,42 @@
...
+    const cronSecret = process.env.CRON_SECRET;
+
+    if (!cronSecret || authHeader !== `Bearer ${cronSecret}`) {
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    const result = await runCalendarSync(db);
+    if (result.failed.length > 0) {
+      console.error("[Cron] calendar feed syncs failed:", result.failed);
+    }
```
positonic, line 35: **[Medium] Sync failures are invisible to operators**

Per-feed failures land in `console.error` here and in `lastSyncError` + a sidebar tooltip; the ICS merge failures in `calendar.getTodayEvents` and `getEventsMultiCalendar` are also bare `console.error`. Nothing reports to Sentry, so a feed stuck in `error` for weeks — or the sweep silently starving (see the `lastSyncedAt` ordering comment) — is undiscoverable in operation. The repo convention for gracefully-handled errors is `reportHandledError` (see `src/server/utils/reportHandledErrorServer.ts`, used by the matrix routers this guard was modeled on).

Suggested fix: `reportHandledError` on cron-level failures and on a feed's transition into `error` state (not on every recurring failure, to avoid noise).

<details>

In `src/app/api/cron/sync-calendars/route.ts` (both catch paths) and in the ICS `.catch` fallbacks in `src/server/api/routers/calendar.ts` (`getTodayEvents` ~line 514, `getEventsMultiCalendar` ~line 852), failures are logged with bare `console.error`. This repo's convention is to route gracefully-handled errors through `reportHandledError` (see `src/server/utils/reportHandledErrorServer.ts` and its call sites in the matrix routers) so they reach Sentry and are findable. Update these paths to use it. To avoid alert noise, report a feed failure only when the feed *transitions* into error state (its previous `syncStatus` wasn't `"error"`), which you can determine in `syncFeed` in `src/server/services/calendar/CalendarSyncService.ts` where the previous feed row is already loaded. Constraints: don't let reporting failures break the sync flow (wrap or fire-and-forget as the existing call sites do); keep the cron response shape unchanged.

</details>

src/server/services/calendar/feedUrlGuard.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Throw unless `rawUrl` is a calendar feed URL this server may safely fetch.
+ *
+ * Resolves DNS, so callers should treat it as a network operation. Called on
+ * `addFeed` and again on every sync fetch (the URL is a stored input, and
+ * DNS answers change).
+ */
+export async function assertSafeFeedUrl(
```
positonic, line 33: **[Medium] Guard skeleton duplicated — this is a line-for-line copy of the Matrix guard**

The extraction moved only `isPrivateAddress`/`defaultResolveHost` into `~/server/utils/privateAddress`; the orchestration — `localhostAllowed`, `isLocalhostName`, protocol check, bracket strip, literal-IP check, resolve-and-judge — is duplicated verbatim from `assertSafeHomeserverUrl` with only error class and message strings differing. The next hardening fix (DNS pinning, a new private range check, the IPv6 normalization suggested elsewhere in this review) now has to land twice, and the copy that gets missed is the vulnerability.

Suggested fix: hoist a shared `assertSafeOutboundUrl(rawUrl, { resolveHost, makeError })` into `privateAddress.ts` and have both guards wrap it with their own error types/messages.

<details>

`src/server/services/calendar/feedUrlGuard.ts` (`assertSafeFeedUrl`, lines ~49–96) duplicates the structure of `assertSafeHomeserverUrl` in `src/server/services/matrix/homeserverUrl.ts` (~lines 45–106): same localhost allowance in non-production, same https-only check, same bracket-stripping + `isIP` literal check, same resolve-and-judge loop — only the error class and message wording differ. Extract the shared skeleton into `src/server/utils/privateAddress.ts` as something like `assertSafeOutboundUrl(rawUrl, { resolveHost, protocolErrorMessage, makeError })`, and reduce both guards to thin wrappers that supply their own error types and user-facing wording. Constraints: keep both existing error classes (`UnsafeFeedUrlError`, `UnsafeHomeserverUrlError`) and their exact messages — callers and tests match on them; keep `feedUrlGuard.test.ts` and the matrix guard's tests green without weakening any assertion; the `ResolveHost` injection point must remain for testability.

</details>

src/app/_components/calendar/CalendarFeedsSection.tsx
```diff
@@ -0,0 +1,295 @@
...
+        notifications.show({
+          title: "Calendar feed added",
+          message: `${feed.name} is now syncing into your calendar.`,
+          color: "blue",
+        });
+      }
+      if (tzData?.timezone == null) {
+        setTzSuggestion(feed.timezone);
+        setTzPromptOpen(true);
+      }
```
positonic, line 98: **[Medium] Double timezone modal on adding the first feed from /calendar**

Adding a first feed here opens this section's `TimezonePromptModal` — but the same success invalidates `listFeeds`, which flips `hasFeeds` → `calendarConnected` in `CalendarPageContent`, whose own prompt condition (`timezone === null && !dismissed`) then also becomes true: two stacked modals. Separately, this section's `onClose` never writes `TZ_PROMPT_DISMISSED_KEY`, so a user who picks "Not now" is immediately re-prompted by the page-level modal.

Suggested fix: make the page-level modal the single owner of the prompt; on addFeed success just stash the X-WR-TIMEZONE suggestion where the page-level prompt can read it (lift the state, or have both write/read the shared sessionStorage dismissal key).

<details>

Two components each own a `TimezonePromptModal`: `src/app/_components/calendar/CalendarFeedsSection.tsx` opens one after a successful `addFeed` (lines ~79–82), and `src/app/_components/calendar/CalendarPageContent.tsx` opens one when a calendar source is connected and `user.getTimezone` returns null (guarded by a `sessionStorage` key `calendar-timezone-prompt-dismissed`). On the /calendar page (where `CalendarFeedsSection` renders inside the sidebar), adding the first feed triggers both: the section's modal opens directly, and the `listFeeds` invalidation makes `CalendarPageContent`'s condition true as well — stacked modals; and dismissing the section's modal doesn't set the sessionStorage key, so the page-level one appears right after. Refactor so exactly one prompt shows: make `CalendarPageContent` the sole owner on /calendar (pass the feed's suggested timezone up or through a shared store), and have `CalendarFeedsSection`'s own prompt (still needed on the settings page, which has no page-level prompt) write the same `calendar-timezone-prompt-dismissed` key on close. Constraints: keep the settings → integrations flow prompting after addFeed; don't prompt users who already have a timezone; keep the X-WR-TIMEZONE prefill behavior.

</details>

src/server/api/routers/calendar.ts
```diff
@@ -357,9 +497,34 @@ export const calendarRouter = createTRPCRouter({
...
-      if (isGoogleCalendarGated(ctx.session.user.email, provider)) return [];
+      const userId = ctx.session.user.id;
+
+      // ICS feed events merge in regardless of provider or Google gating —
+      // every consumer calls this once with the default provider, so this is
+      // the single place today's DB-backed events enter the Today surfaces.
+      const todayStart = new Date();
+      todayStart.setHours(0, 0, 0, 0);
+      const todayEnd = new Date(todayStart);
+      todayEnd.setDate(todayEnd.getDate() + 1);
```
positonic: **[Low] ICS "today" window is server-local (UTC on Vercel) midnight, not the user's day**

`todayStart`/`todayEnd` come from `new Date()` + `setHours(0,0,0,0)` — the server's day, which on Vercel is UTC. For a user in UTC+12, evening events land on the wrong "today". This matches the pre-existing provider convention (`GoogleCalendarService.getTodayEvents` does the same), so it's consistent for V1 — but this PR now collects and validates `User.timezone`, and the schema comment says scheduling interpretation depends on it. Worth a follow-up ticket if not fixed here: compute the window in the user's stored timezone when present.

<details>

In `src/server/api/routers/calendar.ts`, `getTodayEvents` (~lines 505–508) computes the ICS event window as server-local midnight to midnight (`new Date()` + `setHours(0,0,0,0)`), which on Vercel means the UTC day rather than the user's day. The `User.timezone` column (IANA name, validated in `user.updateTimezone`) now exists. When the caller's user has a timezone set, compute today's [start, end) in that zone — e.g. format `new Date()` through `Intl.DateTimeFormat` with `timeZone` to get the local calendar date and offset, and derive the UTC instants for that day's bounds; fall back to the current behavior when timezone is null. Constraints: don't change the provider-event path (`service.getTodayEvents`) in this pass — only the ICS window; keep the response shape identical; add a unit test with a mocked user timezone (e.g. Pacific/Auckland) asserting an event at 22:00 local lands in "today".

</details>

src/server/services/calendar/CalendarSyncService.ts
```diff
@@ -0,0 +1,358 @@
...
+  feedId: string,
+  resolveHost?: ResolveHost,
+): Promise<SyncFeedResult> {
+  const feed = await db.calendarFeed.findUnique({ where: { id: feedId } });
+  if (!feed) return { ok: false, error: "Feed not found" };
+
+  const recordFailure = async (message: string): Promise<SyncFeedResult> => {
+    await db.calendarFeed.update({
+      where: { id: feedId },
+      data: { syncStatus: "error", lastSyncError: message.slice(0, 500) },
```
positonic: **[Low] `lastSyncError` stores raw `error.message` — unguarded channel to the client**

`recordFailure` persists whatever `error.message` says (sliced to 500 chars), and `listFeeds`/`addFeed` return it. All *current* error paths use URL-free messages, but any future throw that embeds the URL (a `new URL(location, currentUrl)` failure on a garbage redirect Location, a library parse error quoting input) would put the decrypted bearer-secret URL into a client-visible field — the exact thing ADR-0057 forbids. Owner-only surface today, so low impact, but the invariant has no enforcement.

Suggested fix: scrub before persisting —

```ts
data: { syncStatus: "error", lastSyncError: message.replaceAll(url, "<feed url>").slice(0, 500) },
```

or map errors to a fixed set of safe strings.

<details>

In `src/server/services/calendar/CalendarSyncService.ts`, `recordFailure` (~line 256) persists raw `error.message` into `CalendarFeed.lastSyncError`, which `listFeeds` and `addFeed` in `src/server/api/routers/calendar.ts` return to the client. ADR-0057 says the feed URL (a bearer secret) must never reach the client. Harden this: before persisting, redact any occurrence of the decrypted feed URL (and its origin/hostname+path) from the message — `recordFailure` is defined inside `syncFeed` after `url` is decrypted, so the value is in scope for most paths; for the decrypt-failure path there is nothing to redact. Keep messages otherwise intact (they're useful in the UI tooltip). Add a unit test that a thrown error whose message contains the URL results in a stored `lastSyncError` without it.

</details>

src/server/api/routers/calendar.ts
```diff
@@ -189,6 +193,142 @@ async function resolveAccount(
...
+  // procedures above — a feed has no account or token behind it. ICS enters
+  // the system here (CRUD + sync) and at the read-path merges only.
+  // ============================================
+
+  addFeed: protectedProcedure
+    .input(
+      z.object({
+        url: z.string().min(1, "Feed URL is required"),
+        name: z.string().trim().min(1).max(100).optional(),
+      }),
```
positonic, line 215: **[Low] No per-user feed cap, no URL length cap**

`url: z.string().min(1)` accepts megabyte-long strings (encrypted and stored), and nothing limits how many feeds one user adds. One user with hundreds of feeds consumes the entire 50-feed cron batch and quietly breaks the ≤30-minute freshness requirement for everyone else.

Suggested fix:
```ts
url: z.string().min(1).max(2048, "Feed URL is too long"),
```
plus a count check in the mutation (e.g. reject past ~20 feeds per user).

<details>

In `src/server/api/routers/calendar.ts`, the `addFeed` mutation (~line 204) validates `url` only with `z.string().min(1)` and never checks how many feeds the user already has. Add `.max(2048)` to the url schema and, inside the mutation before creating, count the caller's existing `calendarFeed` rows and throw a `TRPCError` (`BAD_REQUEST`, message explaining the limit) past a cap of 20. Constraints: add router tests for both rejections in `src/server/api/routers/__tests__/calendarFeeds.test.ts` using the existing `createMockCaller` pattern; keep existing tests green.

</details>

src/app/_components/calendar/CalendarFeedsSection.tsx
```diff
@@ -0,0 +1,295 @@
...
+                  >
+                    Refresh feeds
+                  </Menu.Item>
+                  <Menu.Item
+                    leftSection={<IconUnlink size={14} />}
+                    color="red"
+                    onClick={() => removeFeed.mutate({ feedId: feed.id })}
+                  >
+                    Remove feed
+                  </Menu.Item>
```
positonic, line 252: **[Low] One-click destructive remove; error tooltip unreachable by keyboard**

"Remove feed" fires immediately — deleting the feed and all its events — and since the URL is (correctly, per ADR-0057) never returned to the client, an accidental click means re-hunting for the ICS link. Two smaller UX notes: the per-feed "…" menu's "Refresh feeds" refreshes *all* feeds, which is surprising in a per-feed menu; and the sync-error state is a `Tooltip` on a non-focusable icon, invisible to keyboard/screen-reader users.

Suggested fix: wrap removal in a confirmation modal (the Mantine pattern used elsewhere in the app), and give the error icon a focusable wrapper with an `aria-label` carrying the error text.

<details>

In `src/app/_components/calendar/CalendarFeedsSection.tsx`: (1) the "Remove feed" `Menu.Item` (~line 214) calls `removeFeed.mutate` directly — add a Mantine confirmation step (`modals.openConfirmModal` from `@mantine/modals` if the app has a ModalsProvider, otherwise a small controlled confirm `Modal`) stating that the feed's events will be removed and the URL can't be recovered; (2) the sync-error indicator in `statusFor` (~line 140) is a `Tooltip` around a bare icon — make it keyboard-accessible by wrapping the icon in a focusable element (e.g. `ActionIcon` variant="transparent") with `aria-label={feed.lastSyncError ?? "The last sync failed."}`. Constraints: follow the repo's no-hardcoded-colors rule (semantic Tailwind tokens / Mantine theme colors only); match the existing compact sidebar styling.

</details>

vercel.json
```diff
@@ -37,6 +37,10 @@
     {
       "path": "/api/cron/adr-sync",
       "schedule": "0 * * * *"
+    },
```
positonic, line 40: **[Low] Rider: this PR also activates the Decision Log's `/api/cron/adr-sync` hourly**

The `adr-sync` cron entry is unrelated to calendar feeds and isn't mentioned in the PR description or any commit message. The route exists on main (from the V3 Decision Log PR) but was never registered in `vercel.json`, so merging this PR is what turns that cron on in production for the first time. If intentional, a line in the PR description would keep the deploy diff honest; if not, it should ship separately.

<details>

In `vercel.json`, this PR adds two cron entries: `/api/cron/sync-calendars` (this feature) and `/api/cron/adr-sync` at `0 * * * *` (the Decision Log sync from an earlier PR, whose route exists but was never registered). Confirm with the author whether activating adr-sync is intended here. If not, remove the adr-sync entry from this PR's `vercel.json` change and propose it as its own one-line PR. If yes, update the PR description to mention that this PR activates the ADR-sync cron in production.

</details>

positonic (commented):
Requesting changes for the three High findings — the silent event wipe on non-ICS 200 responses, the cron-sweep starvation via `lastSyncedAt`, and the uncapped aggregate recurrence expansion — plus the feed-scoped unique constraint (Medium) which is cheapest to fix before this migration ships rather than after. The starvation and rate-limit findings share one fix (`lastSyncAttemptAt`). Everything else is non-blocking.

---
## Adversarial review — 3 High / 8 Medium / 5 Low

14 inline comments below; 2 findings with no in-diff anchor follow here. Worst issue: **a feed that 200s with non-ICS content (e.g. an expired Outlook link's HTML sign-in page) silently deletes every synced event and reports `syncStatus: "ok"`** — the inverse of the documented failure contract. Note the starvation and rate-limit findings share one root cause and one fix (`lastSyncAttemptAt`).

Verified sound, for the record: ADR-0057 conformance (encrypted URL never selected client-side, `providerSchema` untouched, ICS only in read-path merges), migration SQL matches schema exactly, delete-and-replace is atomic with failure retention, all-day UTC re-anchoring is correct against node-ical 0.27.1's actual parsing, recurrence test coverage matches the commit claims, ownership scoping on all new mutations, no hardcoded colors.

---

### [Medium] `mastra.getCalendarEvents` (today/upcoming) doesn't see ICS events

The final commit says agent calendar tools see feed events via the `services/index.ts` `getEventsMultiCalendar` merge — but `mastra.getCalendarEvents` (src/server/api/routers/mastra.ts:2252) instantiates `GoogleCalendarService` directly for the `today`/`upcoming` timeframes, so the most-used agent calendar surface still misses feed events. The comment added in `calendar.getTodayEvents` ("every consumer calls this once… the single place today's DB-backed events enter the Today surfaces") is false for this caller. Either merge `listIcsCalendarEvents` there too, or correct the claim and file the follow-up.

### [Medium] `fetchIcsText` — the actual SSRF defense — is untested

The redirect-hop re-validation (a public URL 302ing to `169.254.169.254` is *the* SSRF scenario the design defends against), the streamed 5 MB cap, the timeout, and the Location-less-3xx handling in `CalendarSyncService.fetchIcsText` have no tests. The guard tests only cover `assertSafeFeedUrl` on direct URLs, and the integration tests stub `fetch` with plain 200s. A future "simplification" to `redirect: "follow"` would pass the entire suite while silently removing the defense. Suggest unit tests with a mocked `fetch`: redirect to a private address (expect `UnsafeFeedUrlError`), body stream exceeding 5 MB (expect the cap error and `reader.cancel`), 302 without Location, and >3 hops.

src/server/services/calendar/CalendarSyncService.ts
positonic: ✅ Addressed in 97539054 — a 200 body without BEGIN:VCALENDAR now fails the sync via recordFailure, retaining prior rows; integration regression test covers the HTML-sign-in-page case.

src/server/services/calendar/CalendarSyncService.ts
positonic: ✅ Addressed in 97539054 — new `lastSyncAttemptAt` (migration 2) stamped before any fallible work; the cron sweep and the refreshMyFeeds rate limit both order/judge on it. Integration test pins that failed attempts stamp it while lastSyncedAt stays put.

src/server/services/calendar/CalendarSyncService.ts
positonic: ✅ Addressed in 97539054 — 10k aggregate cap enforced per push (not end-of-parse, so nothing accumulates first), and expandRecurringEvent is wrapped per event so one exploding RRULE skips that event only. Unit tests for both.

prisma/schema.prisma
positonic: ✅ Addressed in 97539054 — unique key is now `[calendarFeedId, externalId, startsAt]` (migration 2). Integration test: same UID+start in two feeds keeps both rows.

src/server/api/routers/calendar.ts
positonic: ✅ Addressed in 8de64bef — rate limit judges lastSyncAttemptAt, only stale feeds re-sync, capped at 10 per invocation. Router tests updated for all three.

src/server/utils/privateAddress.ts
positonic: ✅ Addressed in fdc06877 — IPv6 is normalized to full hextets first; ::ffff:0:0/96 and 64:ff9b::/96 embedded IPv4s are judged with isPrivateIPv4, multicast refused, unparseable refused. Tests for each bypass form plus a public mapped literal staying accepted.

src/app/api/cron/sync-calendars/route.ts
positonic: ✅ Addressed in 97539054 — syncFeed reports via reportHandledErrorServer on the transition into error state (lazy-imported so the pure parse tests stay env-free); cron catch and both router ICS-merge catches report too.

src/server/services/calendar/feedUrlGuard.ts
positonic: ✅ Addressed in fdc06877 — shared `assertSafeOutboundUrl(rawUrl, messages, makeError, resolveHost)` in privateAddress.ts; both guards are now thin wrappers keeping their exact error classes and wording. Both test suites green unweakened.

src/app/_components/calendar/CalendarFeedsSection.tsx
positonic: ✅ Addressed in 8de64bef — CalendarFeedsSection takes an optional `onTimezoneSuggestion`; on /calendar the page owns the single prompt (sidebar hands the X-WR-TIMEZONE up), on settings the section still prompts itself, and every dismissal writes the shared sessionStorage key.

src/server/api/routers/calendar.ts
positonic: ✅ Addressed in 8de64bef — the ICS today-window is computed in User.timezone when set (new pure `todayWindow` helper, DST-aware offset math, provider path untouched). Unit tests: Auckland 22:00-local in today, LA behind-UTC, null-timezone fallback.

src/server/services/calendar/CalendarSyncService.ts
positonic: ✅ Addressed in 97539054 — recordFailure redacts the decrypted URL (`<feed url>`) before persisting; integration test pins it.

src/server/api/routers/calendar.ts
positonic: ✅ Addressed in 8de64bef — `url` capped at 2048 chars, 20 feeds per user; router tests for both rejections.

src/app/_components/calendar/CalendarFeedsSection.tsx
positonic: ✅ Addressed in 8de64bef — removal now goes through openConfirmModal (ModalsProvider is global) with a URL-not-recoverable warning; the error icon is a focusable ActionIcon with the error text as aria-label; menu item renamed 'Refresh all feeds' to match what it does.

vercel.json
positonic: ❌ Declining this one — it's based on a stale diff. `adr-sync` is already registered in vercel.json on origin/main (landed with the V3 Decision Log PR, commit a0a87d5f); this PR's vercel.json diff adds only `/api/cron/sync-calendars`. Verified with `git diff origin/main -- vercel.json`.

PR #595, opened 2026-08-16
fix(cycles): scope cycle date-overlap validation per product

positonic:
## Problem

Creating the first cycle for a product (e.g. Roots) failed with **"Cycle dates overlap with 'Cycle 8'"** even though Cycle 8 belongs to a different product's timeline. Cycles (`List` rows with `listType = SPRINT`) were purely workspace-scoped — there was no product↔cycle relationship at all — so every product in a workspace shared one cycle pool and one overlap constraint.

## Fix

- **Schema**: add `List.productId` (nullable FK to `Product`, indexed, cascade delete). `null` = legacy workspace-shared cycle.
- **Migration backfill**: each existing SPRINT cycle whose tickets all belong to exactly one product is assigned to that product; cycles with no tickets or mixed-product tickets stay shared (`null`).
- **Overlap validation** (`cycle.create` / `cycle.update`): only cycles of the *same product* conflict. Different products — and legacy shared cycles — run independent timelines.
- **Auto-generation / auto-numbering / `cycle.list`**: product-scoped. A product's Cycles tab generates cycles owned by that product; listing returns the product's cycles plus legacy shared ones.
- **Product overview hero**: prefers the product's own current cycle (or one holding its tickets), falls back only to a shared cycle — never another product's.
- **UI**: new-cycle page, Cycles tab, ticket pages, TicketPeek, and retro page pass `productId`, so cycle dropdowns only offer relevant cycles.

## Testing

- 3 new integration tests: overlapping dates allowed across products, still rejected within one product, legacy shared cycle doesn't block a product cycle, foreign-workspace `productId` rejected.
- Full suites pass locally: 2,934 unit + 364 integration (testcontainers ran `migrate deploy`, validating the hand-written migration SQL end to end).
- Migration also applied and verified on the local dev DB.

## Migration note

Contains a DB migration (`20260816090000_add_product_scope_to_cycles`). Independent of the in-flight calendar-feeds branch migration (different tables); merge order doesn't matter. Check the Prisma Migrate Production run after merge.

### DIFF
--- a/prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql
+++ b/prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql
@@ -1,0 +1,26 @@
+-- Cycles (List rows with listType SPRINT) become product-scoped so different
+-- products can run cycles with overlapping dates. NULL productId = legacy
+-- workspace-shared cycle.
+
+-- AlterTable
+ALTER TABLE "List" ADD COLUMN "productId" TEXT;
+
+-- CreateIndex
+CREATE INDEX "List_productId_idx" ON "List"("productId");
+
+-- AddForeignKey
+ALTER TABLE "List" ADD CONSTRAINT "List_productId_fkey" FOREIGN KEY ("productId") REFERENCES "Product"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- Backfill: a cycle whose tickets all belong to exactly one product is that
+-- product's cycle. Cycles with no tickets or mixed-product tickets stay NULL
+-- (workspace-shared) and remain visible on every product's Cycles tab.
+UPDATE "List" l
+SET "productId" = t.pid
+FROM (
+  SELECT "cycleId", MIN("productId") AS pid
+  FROM "Ticket"
+  WHERE "cycleId" IS NOT NULL AND "productId" IS NOT NULL
+  GROUP BY "cycleId"
+  HAVING COUNT(DISTINCT "productId") = 1
+) t
+WHERE l.id = t."cycleId" AND l."listType" = 'SPRINT';
--- a/prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql
+++ b/prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql
@@ -9,11 +9,16 @@ ALTER TABLE "List" ADD COLUMN "productId" TEXT;
 CREATE INDEX "List_productId_idx" ON "List"("productId");
 
 -- AddForeignKey
-ALTER TABLE "List" ADD CONSTRAINT "List_productId_fkey" FOREIGN KEY ("productId") REFERENCES "Product"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+-- SET NULL: deleting a product reverts its cycles to workspace-shared instead
+-- of destroying them (and their SprintMetrics/SprintSnapshot history).
+ALTER TABLE "List" ADD CONSTRAINT "List_productId_fkey" FOREIGN KEY ("productId") REFERENCES "Product"("id") ON DELETE SET NULL ON UPDATE CASCADE;
 
 -- Backfill: a cycle whose tickets all belong to exactly one product is that
 -- product's cycle. Cycles with no tickets or mixed-product tickets stay NULL
 -- (workspace-shared) and remain visible on every product's Cycles tab.
+-- The EXISTS guard skips dirty legacy rows where a ticket's cycle lives in a
+-- different workspace than its product (no DB constraint prevents that), so a
+-- List is never assigned to a product outside its own workspace.
 UPDATE "List" l
 SET "productId" = t.pid
 FROM (
@@ -23,4 +28,8 @@ FROM (
   GROUP BY "cycleId"
   HAVING COUNT(DISTINCT "productId") = 1
 ) t
-WHERE l.id = t."cycleId" AND l."listType" = 'SPRINT';
+WHERE l.id = t."cycleId" AND l."listType" = 'SPRINT'
+  AND EXISTS (
+    SELECT 1 FROM "Product" p
+    WHERE p."id" = t.pid AND p."workspaceId" = l."workspaceId"
+  );
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -712,11 +712,15 @@ model List {
   displayOrder Int @default(0)
 
   workspaceId String
+  // Cycles (listType SPRINT) are scoped to a single product; null = legacy
+  // workspace-shared cycle. Overlap validation only applies within one product.
+  productId   String?
   createdById String
   createdAt   DateTime @default(now())
   updatedAt   DateTime @updatedAt
 
   workspace      Workspace        @relation(fields: [workspaceId], references: [id], onDelete: Cascade)
+  product        Product?         @relation("ProductCycles", fields: [productId], references: [id], onDelete: Cascade)
   createdBy      User             @relation("ListsCreated", fields: [createdById], references: [id])
   actions        ActionList[]
   tickets        Ticket[]         @relation("TicketCycle")
@@ -727,6 +731,7 @@ model List {
 
   @@unique([workspaceId, slug])
   @@index([workspaceId])
+  @@index([productId])
   @@index([createdById])
   @@index([status])
 }
@@ -4160,6 +4165,7 @@ model Product {
   featureDrafts     MeetingFeatureDraft[]
   epics             Epic[]
   repositories      WorkspaceRepository[]
+  cycles            List[]                @relation("ProductCycles")
 
   @@unique([workspaceId, slug])
   @@index([workspaceId])
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -713,14 +713,16 @@ model List {
 
   workspaceId String
   // Cycles (listType SPRINT) are scoped to a single product; null = legacy
-  // workspace-shared cycle. Overlap validation only applies within one product.
+  // workspace-shared cycle. Overlap validation only applies within one
+  // product. On product delete the cycle reverts to workspace-shared
+  // (SetNull) so sprint history survives.
   productId   String?
   createdById String
   createdAt   DateTime @default(now())
   updatedAt   DateTime @updatedAt
 
   workspace      Workspace        @relation(fields: [workspaceId], references: [id], onDelete: Cascade)
-  product        Product?         @relation("ProductCycles", fields: [productId], references: [id], onDelete: Cascade)
+  product        Product?         @relation("ProductCycles", fields: [productId], references: [id], onDelete: SetNull)
   createdBy      User             @relation("ListsCreated", fields: [createdById], references: [id])
   actions        ActionList[]
   tickets        Ticket[]         @relation("TicketCycle")
--- a/src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx
+++ b/src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx
@@ -50,22 +50,33 @@ export default function CyclesListPage() {
   const { workspace, workspaceId } = useWorkspace();
   const [autoCreatePaused, setAutoCreatePaused] = useState(false);
 
+  const { data: product } = api.product.product.getBySlug.useQuery(
+    { workspaceId: workspaceId ?? "", slug: productSlug },
+    { enabled: !!workspaceId && !!productSlug },
+  );
+
   const { data: cycles, isLoading } = api.product.cycle.list.useQuery(
-    { workspaceId: workspaceId ?? "", autoCreate: !autoCreatePaused },
-    { enabled: !!workspaceId },
+    {
+      workspaceId: workspaceId ?? "",
+      productId: product?.id,
+      autoCreate: !autoCreatePaused,
+    },
+    // Wait for the product so auto-generation runs product-scoped, never
+    // against the whole workspace.
+    { enabled: !!workspaceId && !!product?.id },
   );
 
   const utils = api.useUtils();
 
   const deleteCycle = api.product.cycle.delete.useMutation({
     onSuccess: () => {
-      void utils.product.cycle.list.invalidate({ workspaceId: workspaceId ?? "" });
+      void utils.product.cycle.list.invalidate();
     },
   });
 
   const updateCycle = api.product.cycle.update.useMutation({
     onSuccess: () => {
-      void utils.product.cycle.list.invalidate({ workspaceId: workspaceId ?? "" });
+      void utils.product.cycle.list.invalidate();
     },
   });
 
@@ -181,7 +192,7 @@ export default function CyclesListPage() {
         </div>
       )}
 
-      {isLoading ? (
+      {isLoading || !product ? (
         <Stack gap="sm">
           {[1, 2].map((i) => (
             <Skeleton key={i} height={80} />
--- a/src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx
+++ b/src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx
@@ -50,10 +50,11 @@ export default function CyclesListPage() {
   const { workspace, workspaceId } = useWorkspace();
   const [autoCreatePaused, setAutoCreatePaused] = useState(false);
 
-  const { data: product } = api.product.product.getBySlug.useQuery(
-    { workspaceId: workspaceId ?? "", slug: productSlug },
-    { enabled: !!workspaceId && !!productSlug },
-  );
+  const { data: product, isError: isProductError } =
+    api.product.product.getBySlug.useQuery(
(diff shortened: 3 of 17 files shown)


prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql
```diff
@@ -0,0 +1,26 @@
...
+UPDATE "List" l
+SET "productId" = t.pid
+FROM (
+  SELECT "cycleId", MIN("productId") AS pid
+  FROM "Ticket"
+  WHERE "cycleId" IS NOT NULL AND "productId" IS NOT NULL
+  GROUP BY "cycleId"
+  HAVING COUNT(DISTINCT "productId") = 1
+) t
+WHERE l.id = t."cycleId" AND l."listType" = 'SPRINT';
```
positonic: **[Medium] Backfill can link a cycle to a product in a *different workspace* if dirty legacy data exists**

`yourWork.currentCycles` carries a defense-in-depth comment: "no DB constraint forces a ticket's cycle into its own workspace", and the cross-workspace-ref guard only dates from the 2026-08-04 audit — so pre-guard rows where a ticket's `cycleId` points at another workspace's List are plausible. For such a row, this backfill sets `List.productId` to a product in a *different workspace* than the List. That cycle then vanishes from every product tab in its own workspace (it's neither `productId: null` nor any same-workspace product's), and gets cascade-deleted if the foreign product is ever deleted.

Suggested fix — constrain the backfill to same-workspace products:

```suggestion
UPDATE "List" l
SET "productId" = t.pid
FROM (
  SELECT "cycleId", MIN("productId") AS pid
  FROM "Ticket"
  WHERE "cycleId" IS NOT NULL AND "productId" IS NOT NULL
  GROUP BY "cycleId"
  HAVING COUNT(DISTINCT "productId") = 1
) t
WHERE l.id = t."cycleId" AND l."listType" = 'SPRINT'
  AND EXISTS (
    SELECT 1 FROM "Product" p
    WHERE p."id" = t.pid AND p."workspaceId" = l."workspaceId"
  );
```

<details>

In `prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql`, the final UPDATE backfills `List.productId` from the single distinct `Ticket.productId` of each SPRINT cycle's tickets. There is no DB constraint forcing a ticket's `cycleId` to reference a List in the ticket's own workspace (see the comment in `src/server/api/routers/yourWork.ts` around the `cycleTickets` query), so legacy rows could make the backfill point a List at a Product in a different workspace. Add a guard to the UPDATE's WHERE clause: only apply the backfill when a `Product` row exists with `id = t.pid` and `"workspaceId" = l."workspaceId"`. Mismatched cycles must stay `NULL` (workspace-shared). This migration has not shipped to production yet, so edit the SQL file in place rather than adding a new migration — but confirm first (e.g. `npx prisma migrate status` locally) that it hasn't been applied anywhere shared other than the author's local DB; if it has, create a corrective follow-up migration instead. Keep the existing comments accurate.

</details>

prisma/schema.prisma
```diff
@@ -712,11 +712,15 @@ model List {
...
   workspaceId String
+  // Cycles (listType SPRINT) are scoped to a single product; null = legacy
+  // workspace-shared cycle. Overlap validation only applies within one product.
+  productId   String?
   createdById String
   createdAt   DateTime @default(now())
   updatedAt   DateTime @updatedAt
 
   workspace      Workspace        @relation(fields: [workspaceId], references: [id], onDelete: Cascade)
+  product        Product?         @relation("ProductCycles", fields: [productId], references: [id], onDelete: Cascade)
```
positonic: **[Low] Product delete now cascade-deletes backfilled cycles and their sprint history**

`onDelete: Cascade` means deleting a product now destroys its cycles — including *backfilled* legacy cycles that existed before this PR and would previously have survived a product deletion. The cycle's `SprintMetrics`/`SprintSnapshot` rows go with it; retrospectives survive but lose their cycle link (`SetNull`). For cycles the user explicitly created under a product, cascade is arguably right — but for heuristically backfilled ones it's silent history loss.

Consider `onDelete: SetNull` instead: on product delete the cycle reverts to a workspace-shared cycle (`productId: null`), which is exactly the pre-PR semantics and loses nothing. If Cascade is intentional, a note in the migration comment would help the next reader.

<details>

In `prisma/schema.prisma`, the `List.product` relation (relation name "ProductCycles", added in PR 595) uses `onDelete: Cascade`, so deleting a `Product` deletes its cycles (List rows with `listType: SPRINT`), including legacy cycles that a migration backfill assigned to that product. Change the relation to `onDelete: SetNull` so a product deletion reverts its cycles to workspace-shared (`productId: null`) instead of deleting them. `List.productId` is already nullable so this is schema-compatible. Update the corresponding FK in the migration `prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql` (`ON DELETE CASCADE` → `ON DELETE SET NULL`) if that migration has not yet been applied to any shared environment — verify with `npx prisma migrate status`; if it has shipped, create a new migration altering the constraint instead. Do NOT run `prisma db push` or run migrations yourself — ask the user to run `npx prisma migrate dev`. Keep the schema comment on `productId` accurate.

</details>

src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx
```diff
@@ -181,7 +192,7 @@ export default function CyclesListPage() {
         </div>
       )}
 
-      {isLoading ? (
+      {isLoading || !product ? (
```
Comment (positonic) [inline src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx]: **[Low] Permanent skeleton when the product can't be loaded**

`isLoading || !product` shows the skeleton while the product loads — but if `getBySlug` errors or returns nothing (stale bookmark after a product rename/delete), `product` stays undefined forever and the page renders skeletons indefinitely with no error state. The new-cycle page has the sibling problem: the submit button stays disabled with no explanation. Worth branching on the product query's error/empty state and rendering a "product not found" message.

<details>

In `src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx` (PR 595), the render condition `isLoading || !product` shows skeletons while `api.product.product.getBySlug` resolves — but if that query errors or returns no product (deleted/renamed product slug), the page shows skeletons forever. Destructure `isError`/`error` (or check `data === null` after the query settles) from the `getBySlug` useQuery call and render a clear "Product not found" empty state instead of the skeleton in that case. Check how sibling pages under `products/[productSlug]/` handle a missing product and match that pattern if one exists. Apply the same consideration to `cycles/new/page.tsx`, where the Create button is disabled on `!product` with no feedback. Use Mantine components and semantic color classes only (no hardcoded colors — the pre-commit hook rejects them). Keep `npm run check` green.

</details>

src/plugins/product/server/routers/__tests__/cycle-retro.integration.test.ts
```diff
@@ -51,6 +51,98 @@ describe("cycle router", () => {
     expect(updated.achievements).toBe("Shipped email");
   });
 
+  it("allows overlapping dates on cycles of different products", async () => {
```
positonic, line 55: **[Low] Tests cover the create path only — update overlap and auto-generation scoping are untested**

The three new tests all exercise `cycle.create`. Two other paths carry the same new invariant and have none: (1) `cycle.update` date changes, which scope the overlap check via `cycle.productId` — including the subtle `productId: null` Prisma filter for legacy shared cycles; and (2) `ensureUpcomingCycles`, the trickiest code in the PR (legacy shared cycles count as coverage, generated cycles inherit the productId, numbering is per-product). A regression in either would ship silently.

Suggested additions: an update test asserting product A's cycle can be moved onto product B's dates but not onto a sibling A cycle's dates; and a `cycle.list` test with `autoCreate: true` for two products asserting each gets its own generated cycles and legacy shared coverage suppresses generation.

<details>

In `src/plugins/product/server/routers/__tests__/cycle-retro.integration.test.ts` (PR 595 added create-path tests for product-scoped cycle overlap validation), add integration tests for two untested paths in `src/plugins/product/server/routers/cycle.ts`: (1) `cycle.update` — create cycles for two products in one workspace via `caller.product.cycle.create`, then assert updating product A's cycle to dates overlapping product B's cycle succeeds, while updating it to overlap a second cycle of product A itself rejects with /overlap/i; also assert a legacy shared cycle (created via the `createCycle` factory with no productId) can be updated onto dates overlapping a product cycle. (2) `cycle.list` auto-generation — call `caller.product.cycle.list` with `autoCreate: true` and a `productId` for each of two products and assert the generated cycles carry the right `productId` and each product gets its own set; also assert that when legacy shared cycles already cover the lookahead window, no product-scoped cycles are generated. Use the existing factories (`createUser`, `createWorkspace`, `createProduct`, `createCycle` from `src/test/factories/`) and `createTestCaller` from `src/test/trpc-helpers.ts`, matching the style of the tests already in this file. These are `*.integration.test.ts` tests run via `npm run test:integration` (requires Docker). Don't modify router code.

</details>

positonic (commented):
Adversarial review (/pr-review): 6 findings — 1 High, 2 Medium, 3 Low. The product-scoping itself is solid; the issues are at the edges where other code still assumes cycles are workspace-scoped. The one to fix before merge: `workspaceRefs.ts` product-checks epics but not cycles, so every non-UI path can still cross-link tickets to another product's cycle.

---
## PR review — 6 findings (1 High, 2 Medium, 3 Low)

The core change is sound: overlap validation, auto-generation, auto-numbering, and listing are consistently product-scoped, the slug-uniqueness counter loops handle the cross-product `cycle-N` collision, and the migration backfill can't create overlaps that didn't already exist. The gaps are all at the edges of the new invariant — places that still assume cycles are workspace-scoped. Four inline comments below; the two findings without an in-diff anchor follow here.

---

### [High] Ticket↔cycle links can still cross products — the new invariant isn't enforced server-side

`assertWorkspaceScopedRefs` (`src/server/services/access/workspaceRefs.ts`) product-checks epics ("a ticket may not borrow another product's epic") but only workspace-checks cycles. Now that cycles are product-scoped, `ticket.create` / `ticket.update` / `ticket.bulkUpdate` will still happily attach product B's ticket to product A's cycle — the UI dropdowns filter, but the API, SDK, CLI, and mastra agents don't go through the UI. When it happens: product A's cycle detail page lists B's tickets, B's overview hero shows A's cycle (via the `tickets: { some: { productId } }` OR-branch — breaking this PR's own "never another product's cycle" guarantee), and B's ticket-page cycle Select renders blank because the assigned cycle isn't in its filtered options. And it *will* happen, because the SDK's `cycles.current(workspaceId)` picks an arbitrary ACTIVE cycle now that N products each keep one active.

Suggested fix — extend the cycle lookup in `workspaceRefs.ts` to mirror the epic check exactly (a `productId: null` cycle is legacy-shared and exempt, same as a pre-backfill epic):

```ts
if (refs.cycleId) {
  lookups.push(
    db.list
      .findUnique({
        where: { id: refs.cycleId },
        select: { workspaceId: true, productId: true },
      })
      .then((row) => {
        if (row && productId && row.productId && row.productId !== productId) {
          throw new TRPCError({
            code: "NOT_FOUND",
            message: "Cycle not found in this product",
          });
        }
        return { label: "Cycle", refWorkspaceId: row?.workspaceId ?? null };
      }),
  );
}
```

<details>

In `src/server/services/access/workspaceRefs.ts`, the `assertWorkspaceScopedRefs` function validates a ticket's linked references. The epic lookup enforces that an epic with a non-null `productId` must match the pointing ticket's `productId`; the cycle lookup (the `refs.cycleId` block) only checks `workspaceId`. Cycles (List rows, `listType: SPRINT`) now carry a nullable `productId` (PR 595): a cycle with `productId` set belongs to one product, and `productId: null` means a legacy workspace-shared cycle that any product's tickets may join.

Extend the cycle lookup to select `productId` and, when both the function's `productId` parameter and the cycle row's `productId` are non-null and differ, throw `TRPCError({ code: "NOT_FOUND", message: "Cycle not found in this product" })` — exactly mirroring the epic check above it in the same file. Cycles with `productId: null` must remain linkable from any product (do not break existing legacy links). Don't change the function signature. Add an integration test in `src/plugins/product/server/routers/__tests__/` (follow the style of `cycle-retro.integration.test.ts`, using factories from `src/test/factories/`) asserting that `ticket.update` rejects a `cycleId` belonging to a different product of the same workspace, and still accepts a `productId: null` cycle. Keep all existing tests green.

</details>

---

### [Medium] Workspace-wide cycle consumers are left unscoped — agents will pick the wrong product's cycle

Two consumers still assume "one cycle pool per workspace":

1. `mastra.listCycles` (`src/server/api/routers/mastra.ts`) accepts a `productId` but only uses it to *resolve the workspace*, then returns every cycle in it — its doc comment ("Cycles are workspace-scoped") is now stale. An agent asking for product X's cycles gets other products' cycles as valid-looking options.
2. The SDK's `cycles.current(workspaceId)` / `resolve(workspaceId, "current")` return "the single ACTIVE cycle" — but with per-product timelines there are now N ACTIVE cycles and it silently picks the most recently started. Worse, auto-numbering is per-product, so two products can both own a cycle *named* "Cycle 9" in one workspace, making `resolve()`-by-name ambiguous.

Until the High finding (server-side ticket↔cycle product check) lands, both paths let agents/CLI file tickets against another product's cycle with no error.

Suggested fix (this repo's side): in `mastra.listCycles`, when `productId` is provided, filter `where: { workspaceId, listType: "SPRINT", OR: [{ productId }, { productId: null }] }` — same shape as `cycle.list` — and update the stale comment. The SDK needs a follow-up in its own repo (accept `productId` in `list`/`current`/`resolve`).

<details>

In `src/server/api/routers/mastra.ts`, the `listCycles` procedure (search for "listCycles") accepts `{ productId?, workspaceId? }`. When `productId` is given it resolves the product's workspace via `loadProductWithAccess`, but the subsequent `db.list.findMany` filters only on `{ workspaceId, listType: "SPRINT" }`. Since PR 595, cycles carry a nullable `List.productId` (null = legacy workspace-shared cycle). Change the query so that when `input.productId` is provided, the where clause becomes `{ workspaceId, listType: "SPRINT", OR: [{ productId: input.productId }, { productId: null }] }` — matching the scoping used by `cycle.list` in `src/plugins/product/server/routers/cycle.ts`. When only `workspaceId` is provided, keep the current workspace-wide behavior. Also update the block comment above the procedure, which still says "Cycles are workspace-scoped". Don't change the input/output shape, and keep the procedure a pure read (no auto-creation). Note this file can be misdetected as binary by BSD grep — use `grep -a` or `rg` when searching it.

</details>

prisma/migrations/20260816090000_add_product_scope_to_cycles/migration.sql
positonic: ✅ Addressed in c94d0d01 — backfill now guarded with `EXISTS (SELECT 1 FROM "Product" p WHERE p."id" = t.pid AND p."workspaceId" = l."workspaceId")`; cross-workspace dirty rows stay NULL (workspace-shared). Migration edited in place: it had only ever been applied to the local dev DB (verified via `_prisma_migrations`), never to any shared environment.

prisma/schema.prisma
positonic: ✅ Addressed in c94d0d01 — relation changed to `onDelete: SetNull` (schema + migration FK). Deleting a product now reverts its cycles to workspace-shared, preserving SprintMetrics/SprintSnapshot history; the schema comment documents the choice.

src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx
Comment (positonic) [reply src/app/(sidemenu)/w/[workspaceSlug]/products/[productSlug]/cycles/page.tsx]: ✅ Addressed in c94d0d01 — Cycles tab renders a "Product not found" EmptyState when `getBySlug` errors, and the new-cycle page shows an inline error message instead of a silently disabled submit.

src/plugins/product/server/routers/__tests__/cycle-retro.integration.test.ts
positonic: ✅ Addressed in c94d0d01 — added integration tests for the update path (cross-product move allowed, sibling-cycle overlap rejected, legacy null-scope move allowed), per-product auto-generation with legacy-coverage suppression, and ticket→foreign-cycle rejection. Full suite green: 2934 unit + 376 integration.

positonic:
Addressed all 6 review findings in c94d0d01:

- **[High] ticket↔cycle cross-product links**: `assertWorkspaceScopedRefs` now product-checks cycles exactly like epics — `ticket.create`/`update`/`bulkUpdate` reject another product's cycle on every path (API, SDK, CLI, agents), while legacy shared cycles (`productId: null`) stay linkable. Covered by a new integration test.
- **[Medium] `mastra.listCycles`**: when given a `productId` it now filters to `{ productId } OR { productId: null }`, matching `cycle.list`; stale "workspace-scoped" comment rewritten.
- **[Medium] backfill workspace guard** and **[Low] Cascade→SetNull**: fixed in the migration + schema (details in the inline replies).
- **[Low] product-not-found UI** and **[Low] test coverage**: fixed per inline replies.

SDK follow-up (`cycles.current`/`resolve` accepting a productId) is queued as a separate task in the exponential-sdk repo. Full suite green locally: 2934 unit + 376 integration tests.

PR #600, opened 2026-08-17
feat(meetings): extract action items from meeting notes, not just transcript

positonic:
## Problem

"Create Actions" on a meeting ran AI extraction on the transcript alone. `TranscriptionSession.notes` was stored by the Add Meeting modal but never read by `generateDraftActions`, so an explicit numbered **Action Items** list typed into Notes was silently ignored — the drawer filled instead with paraphrased, ASR-noisy transcript inferences (e.g. "Send the baseline document to *Paul*" where the notes correctly said *James*).

## Approach

Notes are the authoritative, human-curated source; the transcript only fills gaps.

- **`ActionExtractionService.extractFromNotes()`** — notes-specific prompt: every item in an explicit action list is extracted near-verbatim (names, product names, URLs preserved), indented sub-bullets become supporting detail rather than separate actions, `"<Name> to <verb>"` sets the assignee, empty list items are skipped. Items carry `From notes: "..."` provenance in context, so the review drawer shows the source distinction with no UI change.
- **Deterministic fallback** (`extractNotesListItems`) — a plain list parser scoped to the "Action Items" section when that heading exists. Used when `OPENAI_API_KEY` is absent, the model call fails, or the model returns nothing: an explicit list can never again extract to zero items.
- **`extractFromTranscript()` gains `excludeActions`** — notes-derived items are listed in the prompt as already captured ("do not return them, nor any rewording") and seed the dedupe set as a backstop. The regex fallback no longer fires when exclusions explain an empty result.
- **`generateDraftActions()`** extracts notes first, passes them as exclusions to the transcript pass, and merges notes-first via `mergeActionItems` (cap 25). Works across all paths: manual meetings, Fireflies summaries, notes-only meetings. A corrupt summary JSON no longer aborts extraction — notes/transcript still produce drafts.

Screenshot associations are unaffected: the merged array is the exact array handed to the processor (index-parallel), and notes items carry no `screenshotRefs`.

## Testing

- 15 new unit tests (`ActionExtractionService.test.ts`) built on a fixture mirroring the real notes that surfaced the bug: pre-heading prose bullet, nested Notion URL sub-bullet, "Zineb to send…" assignee, trailing empty numbered item, post-list prose sections. Pins down section scoping, sub-bullet-as-detail, assignee parsing, notes-first merge/dedup, the already-captured prompt block, and the fallback guard.
- Full unit suite: 282 files / 3058 tests green. Typecheck and targeted lint clean.
- No schema changes — fast-track eligible.

### DIFF
--- a/src/server/services/ActionExtractionService.ts
+++ b/src/server/services/ActionExtractionService.ts
@@ -59,8 +59,10 @@ export function normalizeActionText(text: string): string {
  * Merge action items from multiple sources, earlier sources winning on
  * duplicates. Used to combine notes-derived items (authoritative, human-
  * curated) with transcript-derived ones (inferred). Dedupe here is exact
- * normalized text only — semantic near-duplicates are handled upstream by
- * telling the transcript extraction which actions the notes already yielded.
+ * normalized text only — near-duplicate rewordings are handled upstream:
+ * the AI transcript pass is told which actions the notes already yielded
+ * (excludeActions), and the Fireflies-summary pass (no LLM involved) is
+ * filtered through filterNearDuplicateActions before merging.
  */
 export function mergeActionItems(
   primary: ParsedActionItem[],
@@ -85,6 +87,57 @@ export function mergeActionItems(
   return merged;
 }
 
+const SIMILARITY_STOPWORDS = new Set([
+  "the", "and", "for", "with", "will", "should", "that", "this", "from",
+  "into", "them", "then", "when", "about", "have", "has", "are", "was",
+  "our", "their", "your", "need", "needs",
+]);
+
+function significantTokens(text: string): Set<string> {
+  return new Set(
+    normalizeActionText(text)
+      .split(/[^a-z0-9]+/)
+      .filter((token) => token.length >= 3 && !SIMILARITY_STOPWORDS.has(token))
+  );
+}
+
+/**
+ * Drop candidates that are near-duplicate rewordings of an existing item
+ * ("Zineb will send the baseline doc" vs "Zineb to send baseline document to
+ * James"). Token-overlap against the smaller significant-token set, so a
+ * shorter rewording of a longer item still matches. Deterministic and
+ * dependency-free: it guards the Fireflies-summary path, where no LLM sees
+ * the notes items and exact-match dedupe alone lets rewordings through.
+ */
+export function filterNearDuplicateActions(
+  candidates: ParsedActionItem[],
+  existing: ParsedActionItem[],
+  threshold = 0.6
+): ParsedActionItem[] {
+  const existingTokenSets = existing
+    .map((item) => significantTokens(item.text))
+    .filter((tokens) => tokens.size > 0);
+  if (existingTokenSets.length === 0) {
+    return candidates;
+  }
+
+  return candidates.filter((candidate) => {
+    const tokens = significantTokens(candidate.text);
+    if (tokens.size === 0) {
+      return true;
+    }
+    return !existingTokenSets.some((other) => {
+      let intersection = 0;
+      for (const token of tokens) {
+        if (other.has(token)) {
+          intersection++;
+        }
+      }
+      return intersection / Math.min(tokens.size, other.size) >= threshold;
+    });
+  });
+}
+
 /**
  * Deterministic fallback for written notes: every top-level numbered or
  * bulleted list line is an action item; indented sub-lines are supporting
@@ -94,14 +147,19 @@ export function mergeActionItems(
  */
 export function extractNotesListItems(notesText: string): ParsedActionItem[] {
   const listItemPattern = /^(?:\d+[.)]|[-*•+])\s+(.+)$/;
+  const emptyListItemPattern = /^(?:\d+[.)]|[-*•+])\s*$/;
+  const lineIndent = (rawLine: string): number =>
+    /^\s*/.exec(rawLine)?.[0]?.length ?? 0;
   const items: { text: string; details: string[] }[] = [];
 
   // When the notes have an explicit "Action Items" heading, only the list
   // under it is actions — other bullets (context, observations) are not.
   // Without such a heading, every top-level list item is treated as an action.
+  // Heading markup varies ("Action Items:", "# Action Items", "**Action
+  // Items:**"), so strip the decoration and match the bare words.
   const lines = notesText.split(/\r?\n/);
   const headingIndex = lines.findIndex((line) =>
-    /^#*\s*\**\s*action\s*items?\s*\**\s*:?\s*$/i.test(line.trim())
+    /^action\s*items?$/i.test(line.trim().replace(/^[#*\s]+|[#*:\s]+$/g, ""))
   );
   let scopedLines = lines;
   if (headingIndex !== -1) {
@@ -109,21 +167,38 @@ export function extractNotesListItems(notesText: string): ParsedActionItem[] {
     for (const rawLine of lines.slice(headingIndex + 1)) {
       const trimmed = rawLine.trim();
       const isListLine =
-        listItemPattern.test(trimmed) || /^\d+[.)]\s*$/.test(trimmed);
-      // The section ends at the first non-blank line that isn't part of the
-      // list (a new heading or paragraph).
-      if (trimmed.length > 0 && !isListLine) {
+        listItemPattern.test(trimmed) || emptyListItemPattern.test(trimmed);
+      // The section ends at the first non-indented, non-blank line that isn't
+      // part of the list (a new heading or paragraph). Indented plain lines
+      // are wrapped continuations of the item above, not terminators.
+      if (trimmed.length > 0 && !isListLine && lineIndent(rawLine) === 0) {
         break;
       }
       scopedLines.push(rawLine);
     }
   }
 
+  // Sub-bullets are relative to the list's own base indent, not to column
+  // zero: notes pasted from Notion/Slack often carry a uniform leading indent,
+  // which must not demote every item after the first to a detail.
+  let baseIndent = Infinity;
   for (const rawLine of scopedLines) {
-    const indent = /^\s*/.exec(rawLine)?.[0]?.length ?? 0;
+    if (listItemPattern.test(rawLine.trim())) {
+      baseIndent = Math.min(baseIndent, lineIndent(rawLine));
+    }
+  }
+
+  for (const rawLine of scopedLines) {
+    const indent = lineIndent(rawLine);
     const line = rawLine.trim();
     const match = listItemPattern.exec(line);
+    const lastItem = items[items.length - 1];
+
     if (!match?.[1]) {
+      // An indented plain line under an item is a wrapped continuation of it.
+      if (line.length > 0 && indent > baseIndent && lastItem) {
+        lastItem.details.push(line);
+      }
       continue;
     }
     const text = match[1].trim();
@@ -131,8 +206,7 @@ export function extractNotesListItems(notesText: string): ParsedActionItem[] {
       continue;
     }
 
-    const lastItem = items[items.length - 1];
-    if (indent > 0 && lastItem) {
+    if (indent > baseIndent && lastItem) {
       lastItem.details.push(text);
     } else {
       items.push({ text, details: [] });
@@ -290,32 +364,35 @@ export class ActionExtractionService {
       excludeActions.map((text) => normalizeActionText(text))
     );
     const results: ParsedActionItem[] = [];
+    let anyChunkParsed = false;
 
     for (let i = 0; i < chunks.length; i++) {
       const chunk = chunks[i]!;
       console.log(`[ActionExtraction] Processing chunk ${i + 1}/${chunks.length} (${chunk.length} chars): "${chunk.slice(0, 100)}..."`);
 
-      const response = await model.invoke([
-        new SystemMessage(buildSystemPrompt()),
-        new HumanMessage(buildChunkPrompt(chunk, excludeActions)),
-      ]);
-
-      const rawContent = typeof response.content === "string" ? response.content : "";
-      console.log(`[ActionExtraction] Raw model response: ${rawContent}`);
+      // The invoke itself is inside the try: an OpenAI rate-limit/timeout on
+      // one chunk must degrade to that chunk contributing nothing, not
+      // propagate and discard what the caller already extracted from notes.
       let parsed: z.infer<typeof extractionSchema> | null = null;
-
       try {
+        const response = await model.invoke([
+          new SystemMessage(buildSystemPrompt()),
+          new HumanMessage(buildChunkPrompt(chunk, excludeActions)),
+        ]);
+        const rawContent = typeof response.content === "string" ? response.content : "";
+        console.log(`[ActionExtraction] Raw model response: ${rawContent}`);
         const json = parseJsonFromModelOutput(rawContent);
         parsed = extractionSchema.parse(json);
         console.log(`[ActionExtraction] Parsed ${parsed.actions.length} actions from chunk`);
-      } catch (parseErr) {
-        console.log(`[ActionExtraction] Failed to parse model response: ${parseErr instanceof Error ? parseErr.message : String(parseErr)}`);
+      } catch (chunkErr) {
+        console.log(`[ActionExtraction] Chunk ${i + 1}/${chunks.length} failed: ${chunkErr instanceof Error ? chunkErr.message : String(chunkErr)}`);
         parsed = null;
       }
 
       if (!parsed) {
         continue;
       }
+      anyChunkParsed = true;
 
       for (const action of parsed.actions) {
         const normalized = normalizeActionText(action.text);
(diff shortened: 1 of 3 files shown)


src/server/services/TranscriptionProcessingService.ts
```diff
@@ -134,54 +135,54 @@ export class TranscriptionProcessingService {
...
       }
 
-      if (!processedData) {
-        console.log("[generateDraftActions] No processedData, returning early");
-        result.success = true;
-        return result;
+      if (transcriptItems.length === 0 && transcriptText) {
+        console.log(`[generateDraftActions] No Fireflies actions, using AI extraction on transcript (${transcriptText.length} chars)`);
+        const { numberedText } = numberScreenshotMarkers(transcriptText);
+        transcriptItems = await ActionExtractionService.extractFromTranscript(
```
positonic: **[High] OpenAI throw during transcript pass discards already-extracted notes items**

`extractFromTranscript` wraps only JSON *parsing* in try/catch — the `model.invoke()` call itself (ActionExtractionService.ts:298) is unwrapped, so a rate-limit/timeout/5xx from OpenAI propagates out of this call, lands in `generateDraftActions`' outer catch, and the `notesItems` extracted moments earlier are never merged or created. Result: zero drafts from an explicit Action Items list — exactly the failure this PR exists to fix. (`extractFromNotes` *does* wrap its `invoke` per-chunk, so the two paths are inconsistent.)

Suggested fix — degrade to notes-only instead of aborting:
```typescript
        let extracted: ParsedActionItem[] = [];
        try {
          extracted = await ActionExtractionService.extractFromTranscript(
            screenshots.length > 0 ? numberedText : transcriptText,
            { excludeActions: notesItems.map((item) => item.text) }
          );
        } catch (extractError) {
          console.error("[generateDraftActions] Transcript extraction failed, continuing with notes items:", extractError);
        }
        transcriptItems = extracted;
```
(Or wrap `invoke` per-chunk inside `extractFromTranscript`, mirroring the notes path.)

<details>

In `src/server/services/TranscriptionProcessingService.ts` around line 170, `generateDraftActions` calls `ActionExtractionService.extractFromTranscript(...)` after having already extracted `notesItems` from meeting notes. In `src/server/services/ActionExtractionService.ts` (~line 298), `extractFromTranscript` calls `model.invoke()` with no try/catch — only the subsequent JSON parsing is guarded. If the OpenAI call throws (network error, 429, 5xx), the exception propagates to `generateDraftActions`' outer catch and the notes-derived items are silently lost — the meeting produces zero draft actions even though the notes had an explicit list.

Fix it so a transcript-extraction failure degrades gracefully: notes items must still be merged and created as drafts. Either (a) wrap the `extractFromTranscript` call site in `generateDraftActions` in try/catch and continue with `notesItems` on failure, or (b) wrap the per-chunk `model.invoke()` inside `extractFromTranscript` in try/catch + `continue`, matching how `extractFromNotes` already does it (that method wraps invoke+parse per chunk).

Constraints:
- Don't change `extractFromTranscript`'s public signature.
- Keep all existing tests in `src/server/services/__tests__/ActionExtractionService.test.ts` green.
- Add a regression test: `invokeMock.mockRejectedValueOnce(new Error("rate limited"))` on the transcript pass while notes items exist, asserting notes-derived drafts are still produced (or, if fixing inside `extractFromTranscript`, that it returns rather than throws).

</details>

src/server/services/ActionExtractionService.ts
```diff
@@ -194,7 +347,11 @@ export class ActionExtractionService {
     }
 
     console.log(`[ActionExtraction] AI extraction found ${results.length} items total`);
-    if (results.length === 0) {
+    if (results.length === 0 && excludeActions.length === 0) {
```
positonic: **[Medium] Zero-result fallbacks conflate "model failed" with "model correctly returned nothing"**

Two sides of the same coin:

1. **Transcript side (this line):** with `excludeActions` present, the regex fallback is skipped on zero results. But zero results also happens when *every chunk failed to parse* (model answered prose) — those failures are caught per-chunk with `continue`, so a fully-failed extraction is indistinguishable from "notes covered everything", and the transcript silently contributes nothing.
2. **Notes side (`extractFromNotes`, ~line 449):** the inverse polarity — if the model *correctly* returns `[]` (notes are context bullets with no heading and no actions, which the notes prompt explicitly tells it to skip), the fallback fires anyway and `extractNotesListItems` promotes **every** top-level bullet to a draft action, undoing the model's correct judgment with junk drafts.

Both fix the same way: track whether at least one chunk parsed successfully, and gate the fallbacks on `anyChunkParsed === false` (i.e. fall back only on failure, not on a legitimately empty result).

<details>

In `src/server/services/ActionExtractionService.ts`, both `extractFromTranscript` (~line 350) and `extractFromNotes` (~line 449) decide whether to run a fallback based only on `results.length === 0`, which conflates two different situations: (a) every model chunk failed to parse (a failure — fallback SHOULD run), and (b) the model successfully returned zero actions (a legitimate outcome — fallback should NOT run, especially in `extractFromNotes` where the deterministic parser would then promote every top-level bullet, including non-action context bullets, to an action).

Fix: in both methods, maintain a boolean like `anyChunkParsed`, set true whenever a chunk's response parses successfully (even with zero actions). Then:
- `extractFromTranscript`: run the regex fallback when `results.length === 0 && !anyChunkParsed` (keep also skipping it when `excludeActions` is non-empty AND at least one chunk parsed — the current comment's rationale — but a fully-failed run with exclusions should still fall back).
- `extractFromNotes`: run `extractNotesListItems` only when `!anyChunkParsed`.

Constraints:
- Don't change either method's public signature.
- Keep existing tests green — note `ActionExtractionService.test.ts` has a test "falls back to the deterministic list parser when the model answers prose" (a parse failure: should still fall back) and "does not run the regex fallback when exclusions explain the empty result" (a successful parse of `{"actions":[]}`: should still skip the fallback). Both remain correct under this change.
- Add one test: model successfully returns `{"actions":[]}` for notes without an Action Items heading → `extractFromNotes` returns `[]` (no fallback junk).

</details>

src/server/services/TranscriptionProcessingService.ts
```diff
@@ -134,54 +135,54 @@ export class TranscriptionProcessingService {
...
+          { excludeActions: notesItems.map((item) => item.text) }
+        );
+        console.log(`[generateDraftActions] AI extraction returned ${transcriptItems.length} items`);
+      } else if (transcriptItems.length === 0) {
+        console.log("[generateDraftActions] No Fireflies actions and no transcript text available");
       }
 
+      const processedData = {
+        summary,
+        actionItems: mergeActionItems(notesItems, transcriptItems),
```
positonic, line 215: **[Medium] Fireflies-summary path bypasses the exclusion mechanism — near-duplicate drafts**

When `FirefliesService.parseActionItems(summary)` returns items, the AI transcript pass (and with it the whole `excludeActions` / "already-captured" mechanism) is skipped. `mergeActionItems` then dedupes on **exact normalized text only** — so notes item "Zineb to send baseline document to James" and Fireflies item "Zineb will send the baseline doc" both become drafts. The doc comment on `mergeActionItems` says semantic near-duplicates are "handled upstream by telling the transcript extraction which actions the notes already yielded" — that's true for the AI path but false for the Fireflies path, which is the *common* path for Fireflies-imported meetings with notes.

Suggested direction: when `notesItems.length > 0` and Fireflies items exist, either run a cheap similarity filter over the Fireflies items (token-overlap against notes texts), or at minimum correct the `mergeActionItems` doc comment so the next reader doesn't rely on a guarantee that doesn't hold on this path.

<details>

In `src/server/services/TranscriptionProcessingService.ts` (~lines 154–181, `generateDraftActions`), when a stored Fireflies summary yields action items via `FirefliesService.parseActionItems`, the AI transcript extraction is skipped — so the `excludeActions` mechanism (which tells the model not to re-emit notes-covered actions) never applies. `mergeActionItems` (in `src/server/services/ActionExtractionService.ts`) dedupes only exact normalized text, so semantically-duplicate items (notes: "Zineb to send baseline document to James" vs Fireflies: "Zineb will send the baseline doc") both become draft actions.

Task: reduce near-duplicate drafts on the Fireflies path. A pragmatic approach: before merging, filter `transcriptItems` (Fireflies-derived) against `notesItems` with a lightweight similarity check — e.g. normalized token-set overlap (Jaccard or shared-significant-token ratio) above a threshold like 0.6 drops the Fireflies item. Keep it deterministic and dependency-free. Also update the misleading claim in `mergeActionItems`' doc comment ("semantic near-duplicates are handled upstream...") to reflect the actual behavior.

Constraints:
- Notes items always win — never drop a notes item.
- Keep `mergeActionItems`' exact-match dedupe as-is; add the similarity filter as a separate step or helper so it's independently testable.
- Add unit tests for the similarity filter (a reworded duplicate is dropped; an unrelated item survives).
- Keep all existing tests green.

</details>

src/server/services/ActionExtractionService.ts
```diff
@@ -26,18 +26,130 @@ const extractionSchema = z.object({
...
+    if (!match?.[1]) {
+      continue;
+    }
+    const text = match[1].trim();
+    if (!text) {
+      continue;
+    }
+
+    const lastItem = items[items.length - 1];
+    if (indent > 0 && lastItem) {
```
positonic: **[Medium] Uniformly-indented lists collapse to a single action in the deterministic parser**

`indent > 0` is an *absolute* check. Notes pasted with a uniform leading indent (common when copying from Notion, Slack, or an editor) make every item after the first a "detail" of the first:

```
  - Task A
  - Task B
  - Task C
```

→ one draft action ("Task A") with B and C buried in its context. Since this parser is the safety net for API-key-absent and model-failure cases, a 10-item list degrading to 1 draft nearly re-creates the original bug.

Suggested fix: compute the minimum indent across matched list lines first (the base indent), then treat a line as a sub-bullet only when `indent > baseIndent`.

<details>

In `src/server/services/ActionExtractionService.ts`, `extractNotesListItems` (~lines 122–140) classifies a list line as a sub-bullet (detail of the previous item) when its absolute leading-whitespace length is `> 0`. This breaks for lists where every line carries the same non-zero indent (pasted from Notion/Slack/editors): the first line becomes the only top-level item and all subsequent lines become its details.

Fix: determine the base indent as the minimum indent among lines matching the list-item pattern within the scoped region, then treat a line as a sub-bullet only when its indent is strictly greater than that base indent. Tabs count as whitespace — measure consistently (the current `/^\s*/` length approach is fine as long as both measurements use it).

Constraints:
- Keep the existing exported signature `extractNotesListItems(notesText: string): ParsedActionItem[]`.
- All existing tests in `src/server/services/__tests__/ActionExtractionService.test.ts` must stay green (the fixture's sub-bullet at 3-space indent under 0-indent numbered items must still be treated as detail).
- Add a regression test: a uniformly 2-space-indented bulleted list of 3 items yields 3 actions, and a deeper-indented sub-line under one of them is still detail.

</details>

src/server/services/TranscriptionProcessingService.ts
```diff
@@ -134,54 +135,54 @@ export class TranscriptionProcessingService {
...
-            actionItems,
-            transcriptText,
-          };
+          summary = JSON.parse(transcription.summary) as FirefliesSummary;
+          transcriptItems = FirefliesService.parseActionItems(summary);
+          console.log(`[generateDraftActions] FirefliesService.parseActionItems returned ${transcriptItems.length} items`);
         } catch (parseError) {
+          // Non-fatal: notes and/or raw transcript extraction below can still
+          // produce drafts even when the stored summary JSON is corrupt.
           console.error("[generateDraftActions] Failed to parse transcription summary:", parseError);
```
positonic, line 165: **[Medium] Summary-parse failure now silently swallowed — use `reportHandledErrorServer`**

Making corrupt summary JSON non-fatal is right, but the error also dropped out of `result.errors` — it's now `console.error` only, invisible outside the server logs. A corrupt stored summary is a data-integrity signal worth an alert. This repo's convention for exactly this case is `reportHandledErrorServer` (`src/server/utils/reportHandledErrorServer.ts` — see `CalendarSyncService` for the pattern).

Suggested fix:
```typescript
        } catch (parseError) {
          // Non-fatal: notes and/or raw transcript extraction below can still
          // produce drafts even when the stored summary JSON is corrupt.
          const { reportHandledErrorServer } = await import("~/server/utils/reportHandledErrorServer");
          reportHandledErrorServer(parseError, {
            context: "generateDraftActions: corrupt TranscriptionSession.summary JSON",
            transcriptionId,
          });
        }
```

<details>

In `src/server/services/TranscriptionProcessingService.ts` (~lines 160–164, inside `generateDraftActions`), the catch block for `JSON.parse(transcription.summary)` was changed to be non-fatal (correct) but now only does `console.error`, where it previously pushed onto `result.errors`. Per this repo's convention, gracefully-handled errors must be reported via `reportHandledErrorServer` from `~/server/utils/reportHandledErrorServer` (see `src/server/services/calendar/CalendarSyncService.ts` ~line 312 for the dynamic-import usage pattern) so they reach Sentry instead of vanishing into server logs.

Update the catch block to call `reportHandledErrorServer(parseError, {...})` with useful context (transcriptionId at minimum). Keep the flow non-fatal — do NOT re-add the error to `result.errors` and do not abort extraction. Match the call signature used elsewhere in the codebase. Keep all existing tests green.

</details>

src/server/services/ActionExtractionService.ts
```diff
@@ -26,18 +26,130 @@ const extractionSchema = z.object({
...
+export function extractNotesListItems(notesText: string): ParsedActionItem[] {
+  const listItemPattern = /^(?:\d+[.)]|[-*•+])\s+(.+)$/;
+  const items: { text: string; details: string[] }[] = [];
+
+  // When the notes have an explicit "Action Items" heading, only the list
+  // under it is actions — other bullets (context, observations) are not.
+  // Without such a heading, every top-level list item is treated as an action.
+  const lines = notesText.split(/\r?\n/);
+  const headingIndex = lines.findIndex((line) =>
+    /^#*\s*\**\s*action\s*items?\s*\**\s*:?\s*$/i.test(line.trim())
```
positonic: **[Low] Fallback parser scoping fragility: bold-with-colon heading missed; continuation lines truncate the list**

Two scoping edge cases in the deterministic parser:

1. The heading regex places `:?` *after* the closing `\**`, so `**Action Items**:` matches but the more common markdown form `**Action Items:**` (colon inside the bold) does not — scoping silently disables and pre-heading context bullets become actions.
2. Inside a scoped section, any non-blank non-list line breaks the loop — including an indented *continuation* line of a wrapped item (`1. Do X` / `   because reasons` / `2. Do Y` drops item 2 and everything after).

Suggested fixes: strip `#`, `*`, `:` and whitespace from the trimmed line before matching the heading, and treat an *indented* non-list line as a continuation (append to the previous item's details) rather than a terminator.

<details>

In `src/server/services/ActionExtractionService.ts`, `extractNotesListItems` (~lines 96–120) has two scoping bugs:

1. The heading regex `/^#*\s*\**\s*action\s*items?\s*\**\s*:?\s*$/i` fails on `**Action Items:**` (colon before the closing asterisks — the most common markdown bolding). Simplest robust fix: strip leading/trailing `#`, `*`, `:` and whitespace from the trimmed line first, then test against `/^action\s*items?$/i`.
2. In the section-scoping loop, a non-blank line that isn't a list item ends the section — but an indented plain-text line is usually a wrapped continuation of the previous item, not a new section. Change the terminator check so only *non-indented* non-list lines end the section; indented non-list lines should be kept (and ideally appended to the previous item's details in the main loop below).

Constraints:
- Keep the exported signature unchanged.
- All existing tests must pass, including the one asserting the "Situation Report (Regarding Noah)" heading (non-indented prose) still terminates the section.
- Add tests for both cases: a `**Action Items:**` heading scopes correctly; a wrapped continuation line doesn't swallow subsequent numbered items.

</details>

positonic (commented):
👍

src/server/services/TranscriptionProcessingService.ts
positonic: ✅ Addressed in 43ff3288 — `model.invoke` is now inside the per-chunk try/catch in `extractFromTranscript` (mirroring the notes path), plus a belt-and-braces try/catch at the `generateDraftActions` call site that degrades to notes-only drafts and reports via `reportHandledErrorServer`. Regression test: a rejected invoke no longer throws.

src/server/services/ActionExtractionService.ts
positonic: ✅ Addressed in 43ff3288 — both methods now track `anyChunkParsed` and fall back only when no chunk parsed (extraction failure), never on a legitimately empty parsed result. New test: notes model returning `{"actions":[]}` yields `[]` with no fallback junk.

src/server/services/TranscriptionProcessingService.ts
positonic: ✅ Addressed in 43ff3288 — new `filterNearDuplicateActions` (significant-token overlap ≥ 0.6 against the smaller set, deterministic, dependency-free) filters transcript/Fireflies items against notes items before merging; `mergeActionItems` doc comment corrected. Tests cover reworded-dup dropped / unrelated survives.

src/server/services/ActionExtractionService.ts
positonic: ✅ Addressed in 43ff3288 — base indent is now the minimum indent across matched list lines; sub-bullet iff `indent > baseIndent`. Regression test: uniformly 2-space-indented 3-item list yields 3 actions with a deeper sub-line still detail.

src/server/services/TranscriptionProcessingService.ts
positonic: ✅ Addressed in 43ff3288 — corrupt summary JSON now reported via `reportHandledErrorServer` (dynamic import, `area` + transcriptionId context), still non-fatal.

src/server/services/ActionExtractionService.ts
positonic: ✅ Addressed in 43ff3288 — heading match now strips `#`/`*`/`:` decoration and tests the bare words (so `**Action Items:**` scopes correctly), and indented non-list lines are treated as wrapped continuations (appended to the previous item's detail) rather than section terminators. Tests added for both.

PR #607, opened 2026-08-18
fix(auth): align Google OAuth scopes with Cloud Console for verification

positonic:
## Why

Google's OAuth verification failed (review email, Aug 11) because the scopes the app requests must be a **strict string match** with the scopes registered in the Cloud Console, and they weren't:

| | Scopes |
|---|---|
| Requested by code (before) | `calendar.events`, `calendar.readonly` |
| Configured in Console | `calendar.events`, `calendar.events.readonly` |

Matching the Console verbatim was not an option: `calendar.events.readonly` is a redundant subset of `calendar.events` (violates the least-privilege rule the reviewer cited), and neither scope permits `calendarList.list`, which the multi-calendar sidebar depends on. The correct granular pair, verified against Google's Calendar API reference, is:

```
https://www.googleapis.com/auth/calendar.events
https://www.googleapis.com/auth/calendar.calendarlist.readonly
```

These cover the app's entire Calendar API surface: `events.list`, `events.insert`, `calendarList.list`.

## What changed

- **`src/lib/googleAuth.ts`** (single scope registry): `calendar.readonly` → `calendar.calendarlist.readonly`, with a comment explaining the strict-match constraint.
- **`gmail.readonly` removed everywhere.** The audit found zero Gmail API usage in this repo or the mastra repo (email sync is IMAP/app-password based) — yet the CRM import flow requested this *restricted* scope, which would have failed the next review round and triggered a CASA audit requirement. The `crm` scope set collapses into `contacts`; the import dialog now requests `type=contacts`.
- **CRM connection check** no longer demands the Gmail scope (`hasAllScopes` could never become true again otherwise).
- **Import dialog copy** no longer promises "Gmail – Read email metadata"; the permission list now mirrors the real consent screen.
- **`/privacy`**: new "AI Features and Google User Data" section with the affirmative Limited Use statement required by the Workspace API User Data and Developer Policy.

Existing tester grants keep working: old tokens hold `calendar.readonly` (a superset of the new pair) and connection checks key off `calendar.events`, which is unchanged.

## Verification

- Repo-wide `tsc --noEmit` clean; ESLint clean; all 3065 unit tests pass.
- Live end-to-end check: seeded the dev fixture, minted a session, ran the dev server with the fixture allowlisted in `GOOGLE_OAUTH_TESTER_EMAILS`, and decoded the actual `accounts.google.com` redirect. `type=calendar` requests exactly `openid userinfo.email userinfo.profile calendar.events calendar.calendarlist.readonly`; `type=contacts` adds only `contacts.readonly`; a stale `type=crm` link falls back to the calendar set. No Gmail scope anywhere.
- `/privacy` renders the new Limited Use section.

## After merge (outside the repo)

1. Cloud Console → Data Access: set exactly `calendar.events`, `calendar.calendarlist.readonly` (+ `contacts.readonly` only if submitting CRM import for verification now); **remove `calendar.events.readonly`**.
2. Re-record the demo video against production showing each configured scope in use.
3. Reply to the review email with the AI-provider list (OpenAI API, Anthropic API — neither trains on API data; Mastra is self-hosted orchestration) and point to the privacy-policy statement.

### DIFF
--- a/src/lib/googleAuth.ts
+++ b/src/lib/googleAuth.ts
@@ -11,9 +11,14 @@ import { db } from "~/server/db";
  * Google OAuth scope sets for incremental authorization.
  *
  * We use incremental authorization to minimize permissions requested during onboarding:
- * - "calendar": Only calendar access (sensitive scope, faster Google verification)
- * - "contacts": Calendar + Contacts (sensitive scopes)
- * - "crm": Calendar + Contacts + Gmail (includes restricted scope, requires security audit)
+ * - "calendar": Only calendar access (sensitive scopes)
+ * - "contacts": Calendar + Contacts, for the CRM contact import (sensitive scopes)
+ *
+ * Google's OAuth verification requires a STRICT string match between the
+ * scopes the app requests in its authorization URIs and the scopes registered
+ * on the Cloud Console consent screen (Data Access) — the review fails when
+ * they diverge by even one scope. Any change to these sets must be mirrored
+ * in the Console configuration and re-demonstrated to Google.
  */
 /**
  * Identity scopes requested alongside every set. Without these the
@@ -30,13 +35,16 @@ const GOOGLE_IDENTITY_SCOPES = [
   "https://www.googleapis.com/auth/userinfo.profile",
 ] as const;
 
-// Calendar access. `calendar.events` grants read/write of events on any
-// calendar the user can access, but it does NOT permit `calendarList.list`
-// (listing the user's calendars) — that needs `calendar.readonly`. The
-// multi-calendar sidebar lists calendars per account, so both are required.
+// Calendar access, kept to the narrowest granular scopes that cover the two
+// Calendar API surfaces the app uses. `calendar.events` grants read/write of
+// events on any calendar the user can access (events.list / events.insert),
+// but it does NOT permit `calendarList.list` — the multi-calendar sidebar
+// needs `calendar.calendarlist.readonly` to enumerate the user's calendars.
+// The broad `calendar.readonly` ("see and download any calendar") is
+// deliberately NOT requested: Google's least-privilege review rejected it.
 const GOOGLE_CALENDAR_SCOPES = [
   "https://www.googleapis.com/auth/calendar.events",
-  "https://www.googleapis.com/auth/calendar.readonly",
+  "https://www.googleapis.com/auth/calendar.calendarlist.readonly",
 ] as const;
 
 export const GOOGLE_SCOPE_SETS = {
@@ -49,19 +57,13 @@ export const GOOGLE_SCOPE_SETS = {
     ...GOOGLE_CALENDAR_SCOPES,
     "https://www.googleapis.com/auth/contacts.readonly",
   ],
-  crm: [
-    ...GOOGLE_IDENTITY_SCOPES,
-    ...GOOGLE_CALENDAR_SCOPES,
-    "https://www.googleapis.com/auth/contacts.readonly",
-    "https://www.googleapis.com/auth/gmail.readonly",
-  ],
 } as const;
 
 export type GoogleScopeType = keyof typeof GOOGLE_SCOPE_SETS;
 
 /**
  * Whether this user may use the Google features that depend on scopes Google
- * has not verified yet (calendar, contacts, Gmail).
+ * has not verified yet (calendar, contacts).
  *
  * Verification for those scopes is still pending, so only the accounts
  * registered as test users on the Google Cloud Console project can actually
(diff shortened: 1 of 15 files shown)


src/server/api/routers/crmContact.ts
```diff
@@ -1332,11 +1332,12 @@ export const crmContactRouter = createTRPCRouter({
...
 
-      // Check if account has all required scopes
+      // Check if account has all required scopes. Must stay a subset of
+      // GOOGLE_SCOPE_SETS.contacts — the set the import dialog's connect
+      // button actually requests — or hasAllScopes can never become true.
       const requiredScopes = [
         "https://www.googleapis.com/auth/calendar.events",
         "https://www.googleapis.com/auth/contacts.readonly",
-        "https://www.googleapis.com/auth/gmail.readonly",
       ];
```
positonic: > ✅ Addressed in [`841897dd`](https://github.com/positonic/exponential/pull/607/commits/841897dd) — refactor(auth): reference GOOGLE_SCOPES instead of hardcoded scope strings

---

**[Medium] Scope requirements hardcoded as strings instead of referencing `GOOGLE_SCOPES`**

The new comment says this list "must stay a subset of `GOOGLE_SCOPE_SETS.contacts` … or `hasAllScopes` can never become true" — but that invariant is enforced only by the comment. This exact site is the one that *had* drifted (it demanded the Gmail scope after nothing requested it), and this PR fixes it by hand. `~/lib/googleAuth` already exports `GOOGLE_SCOPES.CALENDAR` and `GOOGLE_SCOPES.CONTACTS` with these exact values, so the duplication can be removed instead of documented.

Suggested fix:
```ts
const requiredScopes = [GOOGLE_SCOPES.CALENDAR, GOOGLE_SCOPES.CONTACTS];
```

<details>

In `src/server/api/routers/crmContact.ts`, inside the `getGoogleConnection` procedure (~line 1338), `requiredScopes` is a hardcoded array of two Google scope URL strings (`.../auth/calendar.events` and `.../auth/contacts.readonly`). The same values are exported from `src/lib/googleAuth.ts` as `GOOGLE_SCOPES.CALENDAR` and `GOOGLE_SCOPES.CONTACTS`. Replace the string literals with those constants (the file already imports from `~/lib/googleAuth` — extend that import). Keep the comment's substance but trim the part that only existed to warn about the duplication. Constraints: don't change the procedure's output shape; keep `npm run check` green; grep the repo for other hardcoded `googleapis.com/auth/` scope strings outside `src/lib/googleAuth.ts` and consolidate any that duplicate exported constants.

</details>

src/lib/googleAuth.ts
```diff
@@ -156,22 +162,14 @@ export async function hasContactsAccess(userId: string): Promise<boolean> {
...
-
 /**
  * Get the OAuth URL for requesting additional scopes
  *
  * @param scopeType - The scope set to request
  * @param returnUrl - Where to redirect after authorization
  */
 export function getGoogleAuthUrl(
-  scopeType: "calendar" | "contacts" | "crm",
+  scopeType: GoogleScopeType,
```
positonic: > ✅ Addressed in [`f2083bad`](https://github.com/positonic/exponential/pull/607/commits/f2083bad) — refactor(auth): move scope registry to client-safe module, type the dialog's OAuth URL

---

**[Low] `getGoogleAuthUrl` is dead code whose URL logic the dialog duplicates**

This PR updates `getGoogleAuthUrl`'s signature, but the function has zero call sites — `ImportDialog.handleConnectGoogle` hand-builds the identical `/api/auth/google-calendar?type=…&returnUrl=…` URL instead. Two copies of the same URL construction is exactly how the stale `type=crm` string survived in the dialog. Either delete the helper, or make the dialog use it so the `GoogleScopeType` union checks the `type` value at compile time. (Caveat: `googleAuth.ts` imports `~/server/db` at module top, so importing it into the client component likely breaks the build — the honest options are "move the helper to a client-safe module" or "delete it.")

<details>

In `src/lib/googleAuth.ts` (~line 171), `getGoogleAuthUrl(scopeType, returnUrl)` builds `/api/auth/google-calendar?type=…&returnUrl=…` but has no callers anywhere in the repo. Meanwhile `src/app/(sidemenu)/w/[workspaceSlug]/crm/contacts/_components/ImportDialog.tsx` (`handleConnectGoogle`, ~line 120) builds the same URL with a template literal. Preferred fix: import `getGoogleAuthUrl` into `ImportDialog.tsx` and use it — but note `googleAuth.ts` imports `~/server/db`, so verify it's safe in a client component first; if the db import makes it server-only, instead move `getGoogleAuthUrl` to a client-safe module or just delete the unused export. Constraints: no behavior change to the redirect URL; keep `npm run check` green; check for other hand-built `/api/auth/google-calendar?type=` URLs (e.g. settings/calendar pages) and align them the same way.

</details>

positonic:
Follow-up in 090c3f6a: `contacts.readonly` is no longer requested anywhere. The Console only registers the two calendar scopes, and since this repo is public Google's reviewers can literally grep it — so the contacts scope set is removed from the request surface entirely (every authorization URI, including stale `type=contacts`/`type=crm` links, now yields exactly `calendar.events` + `calendar.calendarlist.readonly` + identity scopes; verified against a live dev server). The CRM ImportDialog's connect button is replaced by the awaiting-verification notice; accounts that granted the scope before the freeze still pass the check and can keep importing. `GOOGLE_SCOPES.CONTACTS` remains as a check-only constant, and the re-enable path is documented in `googleScopes.ts`.

Console should therefore be: sensitive = `calendar.events` + `calendar.calendarlist.readonly` only (remove `calendar.events.readonly`, do not add contacts); non-sensitive openid/email/profile unchanged.

PR #608, opened 2026-08-18
feat(email): split sign-in code email from post-first-login welcome email

positonic:
## What

- New-user sign-in email is now minimal: subject "Welcome to Exponential — your sign-in code", greeting, "Sign in & start planning — enter this code on the sign-in page:", the code, and the 10-minute expiry. Nothing else, so the code survives preview panes. Workspace-agnostic by design.
- New `sendFirstLoginWelcomeEmail` — the single onboarding email, fired once from `events.createUser` after the first successful sign-in on **any** provider (after pending-invite acceptance). Replaces both the OAuth-only `sendWelcomeEmail` and the long welcome-with-code template.
- Invited users get a variant frame: subject "Welcome to {workspace} on Exponential", opening "{inviter} added you to {workspace} — you're in." (inviter name falls back to email; blank-guarded). Workspace/inviter names are HTML-escaped.
- Deterministic tailoring: the task-layer bullet names the chat tool the invited workspace actually uses — Matrix only when it has an active `matrix-server` integration, "Slack or Matrix" when both, Slack by default (including organic signups).
- Updated go-deeper copy (personal/project goals bullet, single-page tools bullet) and replaced the "What Exponential won't do" section with "AI handles execution. You handle intent. That's the deal."
- CONTEXT.md: new **Welcome email** glossary entry; fixed stale "magic-link signup" wording in the Applicant-account entry.

## Why

An invitee who clicked "log in with email" received 400 words of onboarding pitch with the sign-in code buried in the middle. The code email is a functional, time-critical message; the pitch now arrives once they're actually in, tailored to the workspace they joined.

## Verification

- Full `tsc --noEmit`, ESLint on changed files, and the unit suite (3065 tests) all pass.
- All four email variants rendered via the real senders with Postmark stubbed: code email contains zero onboarding copy; invited/organic frames, Matrix/Slack slot, and no-inviter fallback all verified.


src/server/auth/config.ts
```diff
@@ -428,6 +422,79 @@ export const authConfig = {
...
+                select: { createdBy: { select: { name: true, email: true } } },
+              }),
+              db.integration.findFirst({
+                where: {
+                  provider: "slack",
+                  status: "ACTIVE",
+                  workspaceId: firstAcceptedWorkspaceId,
+                },
+                select: { id: true },
+              }),
```
positonic: **[Medium] Slack detection keys on `Integration.workspaceId`, which Slack integrations never set**

Every Slack `Integration` row is created with `userId` or `teamId` only (see `integration.ts` `createIntegration`, and the slack router) — the workspace association for Slack lives on `SlackChannelConfig.workspaceId`, not on the `Integration`. So this `findFirst` always returns `null`, `chatTools.slack` is always `false`, and the "Slack or Matrix" phrase is unreachable: a workspace using both tools gets a welcome email naming only Matrix. (The pure-Slack case is masked because "Slack" is already the fallback phrase.)

Suggested fix — detect workspace-level Slack via the channel config instead:

```ts
db.slackChannelConfig.findFirst({
  where: { workspaceId: firstAcceptedWorkspaceId, isActive: true },
  select: { id: true },
}),
```

<details>

In `src/server/auth/config.ts` (~lines 453–460, inside `events.createUser`), the welcome-email context resolution checks for a workspace Slack presence with `db.integration.findFirst({ where: { provider: "slack", status: "ACTIVE", workspaceId: firstAcceptedWorkspaceId } })`. This never matches: Slack `Integration` rows are only ever created with `userId` or `teamId` set (see `src/server/api/routers/integration.ts` around line 825), and workspace association for Slack is modeled on `SlackChannelConfig.workspaceId` (unique per workspace). As a result `chatTools.slack` is always `false` and the "Slack or Matrix" branch of `chatToolPhrase` in `sendFirstLoginWelcomeEmail` is dead.

Replace the Slack probe with a lookup that reflects reality, e.g. `db.slackChannelConfig.findFirst({ where: { workspaceId: firstAcceptedWorkspaceId, isActive: true }, select: { id: true } })` (optionally also requiring the linked integration to be `status: "ACTIVE"` via a relation filter).

Constraints:
- Keep the Matrix probe (`MATRIX_SERVER_PROVIDER` + `workspaceId`) unchanged — Matrix servers ARE workspace-scoped integrations.
- Keep all four lookups inside the single `Promise.all`.
- Don't change `sendFirstLoginWelcomeEmail`'s public signature or the `chatToolPhrase` derivation in `src/server/services/EmailService.ts`.
- Keep `npm run check` green.

</details>

src/server/auth/config.ts
```diff
@@ -428,6 +422,79 @@ export const authConfig = {
...
+              matrix: matrixServer !== null,
+            };
+          }
+        }
+      } catch (error) {
+        // Context is flavor, the email is not: fall through and send generic.
+        console.error("[Auth] Failed to resolve welcome email context:", error);
+      }
+
+      sendFirstLoginWelcomeEmail({
```
positonic: **[Medium] The only onboarding email a user ever gets is fire-and-forget in a serverless handler**

This send is deliberately not awaited, so the Postmark call races the tail of the auth request. On Vercel, once the sign-in response is out the function can be frozen mid-flight — and because this fires only from `events.createUser`, there is no second chance: the user permanently never receives the welcome email, with nothing logged. The DB context queries above are all awaited; only the send itself is left dangling.

Awaiting it keeps the failure containment (`.catch` still swallows errors, so a failed send can't block account creation) and costs one Postmark round-trip (~100–300 ms) on a once-per-account event:

```suggestion
      await sendFirstLoginWelcomeEmail({
```

Alternatively, Next 15's `after()` from `next/server` gets it out of the request path without the drop risk.

<details>

In `src/server/auth/config.ts` at line 490 (inside `events.createUser`), `sendFirstLoginWelcomeEmail({...}).catch(...)` is fired without `await`. In a serverless deployment (Vercel), the function can be frozen once the auth response is sent, so the un-awaited Postmark request may be silently dropped — and since this event fires exactly once per user, the welcome email is then lost forever with no retry path.

Change the call to `await sendFirstLoginWelcomeEmail({...}).catch(...)` so the send completes before the event handler returns. Keep the existing `.catch` (a failed send must never block account creation — only log it). Alternatively use `after()` from `next/server` to schedule it post-response if the added latency on first sign-in is a concern.

Constraints:
- Do not let a send failure throw out of `events.createUser`.
- Don't change `sendFirstLoginWelcomeEmail`'s signature.
- Keep `npm run check` green.

</details>

src/server/services/EmailService.ts
```diff
@@ -255,65 +255,174 @@ Didn't request this? You can safely ignore this email.
...
-   * When set, the email shows a **Sign-in code** to type instead of a button to
-   * click (ADR-0056). Used for the email sign-up path, where a link would be
-   * spent by the recipient's mail scanner before they ever saw it. The OAuth
-   * welcome keeps its button — those users are already signed in.
-   */
-  signInCode?: string;
-}): { htmlBody: string; textBody: string } {
-  const { brandColor, appName, appUrl, ctaUrl, ctaText, showExpiration, greeting, signInCode } = options;
-  const dailyPlannerUrl = `${appUrl}/daily-plan`;
+export async function sendWelcomeWithSignInCodeEmail(
```
positonic, line 273: **[Low] Two near-identical sign-in-code templates will drift**

This function is now a ~95% clone of `sendSignInCodeEmail` above — same HTML shell, same code block, same expiry notice, differing only in title/heading/lead-in/subject. The next deliverability tweak (dark-mode metas, code-block styling, footer copy) will land in one and not the other, and nothing will flag the divergence. Consider a small shared `renderSignInCodeEmail({ subject, heading, leadIn })` for just these two siblings — the removed `generateWelcomeEmailContent` went too wide, but these two really are one template with two headings. Non-blocking.

<details>

In `src/server/services/EmailService.ts`, `sendSignInCodeEmail` (returning users) and `sendWelcomeWithSignInCodeEmail` (line 268, new email addresses) are near-identical: both render the same HTML/text shell around `formatSignInCode(code)` and the `SIGN_IN_CODE_TTL_MINUTES` expiry notice, differing only in `<title>`, the `<h1>` heading, the lead-in paragraph, and the subject line. Extract one private helper, e.g. `renderSignInCodeEmail(options: { heading: string; leadInHtml: string; leadInText: string; code: string })` returning `{ htmlBody, textBody }`, and have both senders call it with their own copy and subject.

Constraints:
- Byte-identical rendered output for both emails (diff the before/after output in a scratch script to prove it).
- Don't change either exported function's signature.
- Scope the helper to these two code emails only — do NOT fold `sendFirstLoginWelcomeEmail` or other templates into it.
- Keep `npm run check` green.

</details>

positonic (commented):
Staff-level `/pr-review` pass. Solid split overall — the code-email/welcome-email separation is clean and the escaping/blank-name guards are careful. Two Medium findings worth fixing before merge (one dead detection path, one silent-drop risk on the send), details inline.

---

Adversarial production-risk review (`/pr-review`): **2 Medium, 2 Low** — no Critical/High. Worst finding: the workspace Slack detection queries a field Slack integrations never populate, so the "Slack or Matrix" tailoring is dead code in production. Inline comments carry the details; one general note below.

---

## [Low] No automated coverage for the new variant logic

`sendFirstLoginWelcomeEmail`'s branching (invited vs organic frame, nameless-inviter fallback, blank-name guard, `chatToolPhrase`) and the context resolution in `events.createUser` are verified only manually with Postmark stubbed. That matrix is exactly the kind of thing that regresses silently in a later copy edit — and notably, manual verification passed `chatTools` in directly, which is why the dead Slack probe (inline comment) wasn't caught: the untested seam was the DB-detection layer.

`acceptPendingInvitationsForUser` was extracted from this same event handler specifically to make join-event behavior unit-testable. The welcome-context resolution deserves the same treatment: extract a `resolveWelcomeEmailContext(db, workspaceId, email)` and test it with `mockDeep<PrismaClient>`, plus a small unit test on `sendFirstLoginWelcomeEmail` with `fetch`/Postmark stubbed asserting subject and the four frame variants.

